PR #258, opened 2015-01-16
Renamed filter to postFilter

Kornel:
- a method filter could be misleading

For a query like

    search in "test" filter {
        terms ...
    } limit 10

I would expect a

     "query": {
          "filtered": {
               "filter": {
                    "terms":{

not a post_filter.

Maybe my expectation is wrong, since I'm missing the query part, but either way, a post_filter should be more explicit in my opinion.

What do you think?

### DIFF
--- a/src/main/scala/com/sksamuel/elastic4s/SearchDsl.scala
+++ b/src/main/scala/com/sksamuel/elastic4s/SearchDsl.scala
@@ -104,7 +104,7 @@ class SearchDefinition(indexesTypes: IndexesTypes) {
     this
   }
 
-  def filter(block: => FilterDefinition): SearchDefinition = {
+  def postFilter(block: => FilterDefinition): SearchDefinition = {
     _builder.setPostFilter(block.builder)
     this
   }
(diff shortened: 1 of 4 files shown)


sksamuel:
I see what you're saying, but if we're doing to do this, it needs to be done in a backwards compatible way, so changing the keyword isn't gonna work. Also I think we could re-tweak the query part to have the filters nested inside perhaps? Then you can do postFilter as you wish, and filter as it stands would remain but deprecated, and the new non-deprecated way of doing a filter would be inside a new query block. Or something. WDYT @fehmicansaglam  too.

Kornel:
Backwards compatibility is important, this time I thought this shouldn't stay this way but what you're saying makes sense - let's make it deprecated and maybe nest inside a filter? This would better reflect the structure of the expected json

sksamuel:
I've merged your PR and added filter as a synonym of postFilter so old code still works.

Kornel:
Thanks for that. I'll try to find some time to nest the filter element soon :)

PR #265, opened 2015-01-30
Now delimiter and replacement PathHierarchyTokenizer parameters are pass...

algd:
...ed to XContentBuilder as String (as expected), not as Number.

The generated object was:
"testTokenizer": {
                    "type": "path_hierarchy",
                    "delimiter": 47,
                    "replacement": 47
                }

With this change, now it is:
"testTokenizer": {
                    ""type": "path_hierarchy",
                    "delimiter": "/",
                    "replacement": "/"
                }

The first object was rejected by elasticsearch with:
org.elasticsearch.ElasticsearchIllegalArgumentException: delimiter can only be a one char value

With the second it works.
Thanks, hope it helps some people :)


sksamuel:
Thanks!

PR #266, opened 2015-02-01
abstract out ElasticSugar.blockUntil so it can be reused for other predicates

mpollmeier:
ElasticSugar used to have only `blockUntilCount`, now we have `blockUntil which takes a message and a predicate` and `blockUntilCount` which uses the former.
This way we can reuse it to assert other things than counts.

### DIFF
--- a/src/test/scala/com/sksamuel/elastic4s/ElasticSugar.scala
+++ b/src/test/scala/com/sksamuel/elastic4s/ElasticSugar.scala
@@ -47,26 +47,28 @@ trait ElasticSugar extends BeforeAndAfterAll with Logging {
     listener.actionGet()
   }
 
-  def blockUntilCount(expected: Long,
-                      index: String,
-                      types: String*) {
-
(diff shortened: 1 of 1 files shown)


sksamuel:
Great idea thanks. Although I could further refine it to make it use a call by name block. WDYT.

mpollmeier:
Hmm, doesn't it already take a call by name argument? 
the method signature is `def blockUntil(explain: String)(predicate: () ⇒ Boolean): Unit` and according to http://stackoverflow.com/questions/13337338/call-by-name-vs-call-by-value-in-scala-clarification-needed this is the definition of cbn / cbv:
```
def callByValue(x: Int) = {
  println("x1=" + x)
  println("x2=" + x)
}

def callByName(x: => Int) = {
  println("x1=" + x)
  println("x2=" + x)
}
```

sksamuel:
It's essentially the same thing, but it can be

blockUntil(explain:String)(thunk : => Boolean)

Then we can write

blockUntil("boom") {
 // code here does not need () =>
}

Issue #267, opened 2015-02-02
Lack of support for `SpanNearQuery`


XuKaiqiang:
There is no mapping from `spanNearQuery` in ElasticSearch QueryBuilder to SpanNearQueryDefinition in this project.

http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/query-dsl-span-near-query.html

PR #268, opened 2015-02-02
added spanNearQuery

XuKaiqiang:
See https://github.com/sksamuel/elastic4s/issues/267


sksamuel:
Nice. Should it extend spanquerydef rather than just querydef ?

XuKaiqiang:
Good catch.

sksamuel:
Wicked thank you.

PR #271, opened 2015-02-09
Change ElasticsearchUri to ElasticsearchClientUri on @deprecated

JD557:
The deprecated messages states "For multiple hosts, prefer the methods that use ElasticsearchUri", while the class name is "ElasticsearchClientUri".


sksamuel:
Good spot thanks.

PR #282, opened 2015-03-04
FIX #280 Extending documentation

### LINKED ISSUE #280: Geo Shape Type - index properties
I'm trying to create a index with a field typed to `GeoShapeType`. The code (thanks for a really smooth API here) looks like this

```scala
client.execute {
    create index MEDIAQ_INDEX mappings {
        MEDIAQ_TYPE as (
            "videoId" typed StringType,
            "trajectory" typed GeoShapeType
        )
    }
}
```

Now I want to add `precision` and the `tree` field to the properties. How can I achieve this or is this currently not supported? [Elasticsearch documentation](http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/mapping-geo-shape-type.html#_example_4)

```json
{
    "properties": {
        "location": {
            "type": "geo_shape",
            "tree": "quadtree",
            "precision": "1m"
        }
    }
}
```

muuki88:
Great answers should not be forgotten :)

### DIFF
--- a/guide/createindex.md
+++ b/guide/createindex.md
@@ -93,6 +93,33 @@ client.execute {
 
 There are many options that can be specified a field or type. The [reference](#create_index_reference) section below describes how many of these features can be used. For a full list, see the methods in the DSL or see the official documentation on [mapping](http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/mapping.html) (the DSL keywords will be the same or very close to the name of the optional in the REST API).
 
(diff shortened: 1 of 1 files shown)


sksamuel:
Thank you for doing this, but its GeoShapeType that supports those attributes not GeoPointType. Can you fix?

muuki88:
Created own section.

sksamuel:
Great thanks.

PR #283, opened 2015-03-05
Add ShingleTokenFilter support

nonontb:
Create the case class ShingleTokenFilter.
Add some tests in AnalyzerTest. Those are still incomplete


nonontb:
Test completed. Is it ok for you ?

Tell me if you need something more

Build has failed from another change

sksamuel:
Thanks!

PR #284, opened 2015-03-08
Weight parameter per function in function score

petro-rudenko:
[By es1.4](https://github.com/elasticsearch/elasticsearch/commit/c5ff70bf430a81702c6e429942a8a56d0287f468):
Weights can be defined per function like this:

```
"function_score": {
    "functions": [
        {
            "filter": {},
            "FUNCTION": {},
            "weight": number
        }
        ...
```
If `weight` is given without `FUNCTION` then `weight` behaves like `boost_factor`.
This commit deprecates `boost_factor`.

### DIFF
--- a/src/main/scala/com/sksamuel/elastic4s/scorers.scala
+++ b/src/main/scala/com/sksamuel/elastic4s/scorers.scala
@@ -41,8 +41,8 @@ trait ScoreDefinition[T] {
     this._filter = Option(filter)
     this.asInstanceOf[T]
   }
-  def weight(weight: Double): T = {
-    builder.setWeight(weight.toFloat)
+  def weight(boost: Double): T = {
+    builder.setWeight(boost.toFloat)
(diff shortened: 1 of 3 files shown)


sksamuel:
If we're going to deprecate boost_factor then we should add a deprecation warning. Also I wouldn't want to remove the tests, lets keep them in.

sksamuel:
The test I mean is this one https://github.com/petro-rudenko/elastic4s/commit/7e458010033a6a921ab158b5e14b286869ed47d2#diff-7b1720b86dd8820ab98e7065bb0f2e96L628

petro-rudenko:
boost_factor is generated by FactorScoreDefinition [which is deprecated](https://github.com/sksamuel/elastic4s/blob/master/src/main/scala/com/sksamuel/elastic4s/scorers.scala#L28)

sksamuel:
Thanks.

PR #290, opened 2015-03-23
Make the library compatible with ES 1.5

Astrac:
With this PR it is possible to use elastic4s when running ES 1.5. There were two minor issues that were caused by the removal of the methods `waitForMerge` and `exclude`. These are relevant commits in the ES repository:

https://github.com/elastic/elasticsearch/commit/a29b4a800dbac84a4aa89b19e0a4100e8d8507ce

https://github.com/elastic/elasticsearch/commit/f735baf306571ec9949514a14828f18058972044


sksamuel:
Great work thanks

PR #296, opened 2015-04-12
AttachmentType to support AttributeFields

k4200:
We can specify `fields` for `attachment` type according to the doc:
https://github.com/elastic/elasticsearch-mapper-attachments


sksamuel:
Thanks!

PR #297, opened 2015-04-13
Change release version in dependency from 0.90.13.10 -> 0.90.13.2

mmilewski:
I don't know where 0.90.13.10 is coming from but `sbt compile` fails with com.sksamuel.elastic4s#elastic4s_2.11;0.90.13.10: not found.
I found here http://search.maven.org/#search%7Cgav%7C2%7Cg%3A%22com.sksamuel.elastic4s%22%20AND%20a%3A%22elastic4s_2.11%22 that there exist version 0.90.13.2 and it actually resolves


sksamuel:
Thanks!

PR #298, opened 2015-04-15
add support for getting snapshot info

bitsofinfo:
added support for getting snapshot info


sksamuel:
This is a great addition, but I don't think we need to add methods which are immediately deprecated. We can just skip those two, and people can use the intended way. What do you think

bitsofinfo:
Looks like its already merged, I don't think it hurts to retain the deprecated syntax. I'll have more PR's coming I imagine.

sksamuel:
Should be in 1.5.5 ?

PR #303, opened 2015-04-29
Add timeout flag to cluster health request

stringbean:
Extends the `clusterHealth` request so you can set a timeout:

```scala
client.execute {
  clusterHealth timeout "2s"
}
``


sksamuel:
Great thanks!

PR #333, opened 2015-06-16
Add docValuesFormat to timestamp mapping

l15k4:
Hi,

as `doc_values` format will be the default in 2.0 for NotAnalyzed fields which `DateType` is, I find it important to have a chance to set it for `_timestamp` too. Also in huge datasets `_timestamp` is n1 field to sort by and it is not possible without `doc_values` format... WDYT ?


sksamuel:
Great thanks

PR #334, opened 2015-06-16
Add indices option to mapping, count and search dsl

Jarlakxen:
The indices option allow us to control how to deal with unavailable concrete indices (closed or missing), how wildcard expressions are expanded to actual indices (all, closed or open indices) and how to deal with wildcard expressions that resolve to no indices.


sksamuel:
Fantastic thanks.

PR #341, opened 2015-06-23
Add context queries to suggestions

lenniboy:
Context queries allow suggestions to be filtered in a limited way. Documentation is here: https://www.elastic.co/guide/en/elasticsearch/reference/current/suggester-context.html

I noticed that they are currently not supported, so I added them.

If there is anything that needs cleaning up, please let me know.


lenniboy:
I've added the override. If there is anything more you'd like me to change, just let me know.

Thanks for all your work.

sksamuel:
Great PR thanks.

PR #342, opened 2015-06-24
Enhance multiget DSL to include get options

stringbean:
This allows you to set `version`, `routing` and `fields` on multiget requests.


sksamuel:
Lovely thanks.

PR #344, opened 2015-06-25
Added IterableSearch for iterating search results/search hits

aaronp:
Let me know what you think about putting this into its own trait.

Initially it was totally separate from ElasticClient, but then I made elastic client mix it in.

There are many ways to slice it, of which this is just one, and it may go against the code style you want to keep


sksamuel:
I could not get this build to pass, but I know its just travis being shit.

PR #346, opened 2015-06-30
Support explicit field types in the update dsl

olger:
Hi Stephen, 

Your elastic4s library has been in use within our projects for quite a while and really love the way it works. 
I am using explicit types to render certain json field (spray json / json4s) in a proper way in the elasticsearch source. 
In order to update these fields, I need the support for explicit fields in the updateDSL. This pull request adds that support in 2 ways. 

First is to update the doc with a FieldValue that is not using the FieldsMapper, second is via the FieldsMapper (docAsUpsert ("captain" -> SimpleFieldValue("james", "kirk")) 

Tests and documentation is added for the addition. 

Kind regards, 

Olger

### DIFF
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/UpdateDsl.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/UpdateDsl.scala
@@ -61,6 +61,11 @@ class UpdateDefinition(indexesTypes: IndexesTypes, id: String)
     _builder.setDocAsUpsert(true)
     doc(map)
   }
+  def docAsUpsert(value: FieldValue): this.type = {
+    _builder.setDocAsUpsert(true)
+    doc(value)
+  }
+
   def doc(fields: (String, Any)*): this.type = doc(fields.toMap)
   def doc(iterable: Iterable[(String, Any)]): this.type = doc(iterable.toMap)
   def doc(map: Map[String, Any]): this.type = {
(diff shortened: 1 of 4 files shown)


sksamuel:
PR looks fine, but can you explain the rationale to me again, as I didn't quite follow your explanation. Are you wanting more fine grained control over how the fields are indexed, than what the field mapper provides?

olger:
Indexing is fine with the [explicit field values](https://github.com/sksamuel/elastic4s/blob/master/guide/index.md#indexing-with-explicit-fields)

When this explicit field value is a custom field value (lets say CustomDateFieldValue("post_date", new Date())) and I like to update that, I need update to support those explicit field values. 

Hope this brings more clarity.

sksamuel:
Yes that makes sense thanks @olger . Also if you have other implementations of FieldValue that you think might be useful, then do free feel to contribute them to the mapper as well.

olger:
Thanks for merging. 

Most of the times I use the FieldValue for writing JSON based types (json4s / spray-json) in elastic search. Without the custom type they end up as a string that contains the json. 
Added those types would get dependencies on those json libraries. (I am willing to contribute them if you wish, they are not really complex though)

sksamuel:
You could do what I did for other parts of the codebase, and introduce a
typeclass used by a field mapper. Then inside elastic-jackson module,
provide a general purpose one using jackson. People can easily write their
own using json4s or spray if they had that need.

Have you seen Indexable in the code base?
https://github.com/sksamuel/elastic4s#indexing-from-classes

olger:
Intresting, I was not aware of this option. 

I’ll have a look and see what I can do. 




Thanks for the pointer, 




Olger

sksamuel:
It's relatively new (last 2 months maybe?) but I think its a good way of indexing your domain objects.

PR #347, opened 2015-06-30
made searches a val so multi-search definitions can be deconstructed

aaronp:
Handy for when you need to get the definitions back out from the multi-search definition,
like when you're pairing them up with the results.


sksamuel:
Thanks!

PR #353, opened 2015-07-10
Validating typeclass for hit deserialization

jeantil:
Introduces a 2-level type class for deserializing Hits 

HitRead reads a RichSearchHit into a type T
HitFieldRead reads a RichSearchHitField into a type T

An implicit conversion is provided to automatically create a HitFieldRead if a HitRead is available for the type. 

A test case demonstrating usage is provided in RichSearchResponseTest, in include basic type conversion, sequence conversion and nested object conversion

### DIFF
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/Reader.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/Reader.scala
@@ -1,5 +1,10 @@
 package com.sksamuel.elastic4s
 
+import org.scalactic._
+
+import scala.collection.{Seq, generic}
+import scala.reflect.ClassTag
+
 @deprecated("use HitAs, this reader trait has a broken contravariance implementation", "1.6.1")
 trait Reader[-U] {
   def read[T <: U : Manifest](json: String): T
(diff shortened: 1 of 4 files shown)


sksamuel:
It looks good on first glance but I want to look over it properly make sure I understand it if that's ok.

jeantil:
sure take your time :)

elastic4s-core/src/main/scala/com/sksamuel/elastic4s/Reader.scala
```diff
@@ -7,4 +12,94 @@ trait Reader[-U] {
...
+
+trait HitFieldRead[T] {
+  def as(prefix:String)(hit: RichSearchHitField): T Or Every[ErrorMessage]
+  def as(hit: RichSearchHitField): T Or Every[ErrorMessage]=as("")(hit)
+}
+
+trait LPHitFieldReads{
+  implicit def hitRead2HitFieldRead[T](implicit hitRead:HitRead[T]) =new HitFieldRead[T] {
+    import scala.collection.JavaConverters._
+    override def as(prefix: String)(hit: RichSearchHitField): Or[T, Every[ErrorMessage]] = hit match {
```
martin-g: All other methods in the PR use `T Or Every[ErrorMessage]`. Just this method seems to be inconsistent.

elastic4s-core/src/main/scala/com/sksamuel/elastic4s/Reader.scala
```diff
@@ -7,4 +12,94 @@ trait Reader[-U] {
...
+      case MissingRichSearchField(name)=> Bad(One(s"$prefix $name field is missing"))
+      case field=> field.value[Any] match {
+        case map:java.util.Map[_,_] => hitRead.as(RichMapSearchHit(map.asInstanceOf[java.util.Map[String,Any]].asScala.toMap))
+        case other => Bad(One(s"$prefix ${hit.name} expected a object (Map[String,Any]) got $other"))
```
martin-g: 1. aN object
2. put a comma before `got`. I think it will be easier to read

jeantil: Hello,

I won't be able to update the PR until September but I'll do it if it is
still there when I get back.

Cheers
Jean

Le jeu. 6 août 2015 22:20, Martin Grigorov <notifications@github.com> a
écrit :

sksamuel:
I'm ready to get merging this into 2.0.0 now. One thing I'm not sure on is the `HitFieldRead` construct. Why do we need this? The reason I ask is that I always envisaged some trait, call it `HitRead` like here, which is responsible for saying, given a RichSearchHit I'll give back to you a T. Or in this case, a `Or[T, ErrorMessage]`. I assumed that the implementation of this would delegate back to some json encoder like json4s or jackson.

sksamuel:
Upon delving into the code a bit more, I can see the reasons behind it, to achieve per-field error checking.

sksamuel:
I've started merging this into master by hand, so I can modify a few bits, and since there are conflicts. I'll close once I've completed that.

jeantil:
cool ! thanks

Issue #358, opened 2015-07-17
Streaming more than 10 results may not complete


dspasojevic:
Hi,

I'm investigating an issue we are seeing where streamed results never complete.

I've run `ScrollPublisherIntegrationTest` locally, and it seems to have the same issue. Many more results than expected are returned. Putting some debugging in `com.sksamuel.elastic4s.streams.PublishActor#PublishActor`, I see that the scroll id is flipping between null, not-null, null, etc (see the output at the bottom).

I think that this is because no keep alive is specified on the PublishActor's scroll request. That seems to result in `org.elasticsearch.action.search.SearchScrollRequest#writeTo` (v1.6.0) setting the scroll flag to false. `SearchScrollRequest` requires both a keep alive and a scroll id to perform the scroll and return a new scroll id.

Is this a bug in elastic4s? If so, I'm happy to submit a pull request that sets the keep alive to a value supplied to the subscriber. I think that checking the result of the .await in the tests is probably also worth doing.

Thanks,
-Dan

Output:

[...]

PR #359, opened 2015-07-17
Set keepalive on scroll requests.

dspasojevic:
Pulls the keepalive setting out of the original search request and supplies to the scroll request.

See #358 for discussion.


sksamuel:
Awesome thanks

PR #364, opened 2015-07-29
Add periodic forced bulk indexing to BulkIndexingSubscriber

franbh:
Introduces optional periodic forced bulk indexing to BulkIndexingSubscriber. This ensures that all elements are indexed, even if the last batch size is lower than the specified batch size. This is important when the publisher will never complete (for example a Kafka consumer.)

A new test case demonstrating this new feature has been added to BulkIndexingSubscriberIntegrationTest.


juanjovazquez:
+1

sksamuel:
This is great thanks

PR #369, opened 2015-08-05
Fix refresh index DSL

### LINKED ISSUE #368: "refresh index" creates RecoveryRequest instead of RefreshRequest
"refresh index" command actually calls "recover index" operation

https://github.com/sksamuel/elastic4s/blob/master/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/ElasticDsl.scala#L487

martin-g:
Fixes #368


martin-g:
I'm looking into the build failure.

sksamuel:
Brill thanks.

PR #372, opened 2015-08-06
bugfix: Proper creation of synonyms filter.

andrestc:
It was creating one synonym field for each member in the set. ES would ignore all of them except the last one.


andrestc:
It would be great if a new version could be release with this fix because we need this feature in production. =)

Thanks a lot for the great work!

sksamuel:
Do you want 1.5.x release or 1.6.x?

andrestc:
Could it be a 1.7.x? Im using 1.7.0 already.

Thanks a lot.

PR #373, opened 2015-08-06
Fixes #371 - Delegate the call to #templates(DynamicTemplateDefinition*)

### LINKED ISSUE #371: StackOverflowError in MappingDefinition.templates
When using a `templates` or `dynamicTemplates` with multiple items you get a StackOverflowError.

This makes sense when looking at https://github.com/sksamuel/elastic4s/blob/1.7.0/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/mappings/MappingDefinition.scala#L141 but is not the desired behavior i think.

Testcase:
```scala
client.execute {
  create template "foobar" pattern "logstash-*" mappings (
    dynamicTemplates(
      template("string_foo") matching "*foo" matchMappingType "string" mapping {
        field typed StringType index NotAnalyzed
      },
      template("string_bar") matching "*bar" matchMappingType "string" mapping {
        field typed StringType index NotAnalyzed
      }
    )
  )
}
```

martin-g:


sksamuel:
Great thanks.

PR #379, opened 2015-08-08
Helper test method to block until a document has an specific version.

andrestc:
Useful for integration tests that rely on updating a document.


sksamuel:
Great thanks,

PR #384, opened 2015-08-20
Fix the name of the index

ExNexu:
It seems `places` is the index and not `cities`. Code: https://github.com/sksamuel/elastic4s/blob/master/elastic4s-core%2Fsrc%2Fmain%2Fscala%2Fcom%2Fsksamuel%2Felastic4s%2FDeleteDsl.scala#L24

### DIFF
--- a/guide/delete.md
+++ b/guide/delete.md
@@ -21,7 +21,7 @@ or
 The syntax is quite SQL like. Of course we don't always want to delete by id, we might want to do a delete based on
 some criteria. Enter the delete by query request.
 
-Let's delete all London's from the cities index.
+Let's delete all London's from the places index.
 
 ```scala
   client.execute {


sksamuel:
The docs are right actually, the overloaded form is called. The type is cities and index places.

ExNexu:
Hey @sksamuel please look again.

The docs say:
`Let's delete all London's from the cities index.`
(https://github.com/sksamuel/elastic4s/blob/master/guide/delete.md)

You're saying:
`The type is cities and index places.`

My commit changes the docs to:
`Let's delete all London's from the places index.`

Thank you.

sksamuel:
Yes I see my bad sorry @ExNexu

ExNexu:
No worries! Thanks!

PR #403, opened 2015-08-22
Fix #401: Checking if SearchDefinition has Scroll to be used by ScrollPublisher.

### LINKED ISSUE #401: If publisher query doesn't have the scroll parameter, throw an exception
Before creating ScrollPublisher, this method should check whether SearchDefinition has a scroll: https://github.com/sksamuel/elastic4s/blob/master/elastic4s-streams/src/main/scala/com/sksamuel/elastic4s/streams/ReactiveElastic.scala#L33

Without a scroll, publisher will be unbound and endlessly return first batch. Check might help devs to avoid getting wrong results.

andrestc:


sksamuel:
Brilliant thanks.

PR #409, opened 2015-08-24
Fixes #408

### LINKED ISSUE #408: Mapping: Source Exclusion directive is not managed by MappingDefinition DSL
When you want to exclude only specifics fields in the source document you may use fine grained exclusion rules as showed in https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-source-field.html#include-exclude .
It is especially useful when you want to index binaries and keeping fields in types without holding the heavy Base64 payload.
If you use com.sksamuel.elastic4s.mappings.MappingDefinition#source , none of the source fields are kept so you can only retrieve the document id when searching.

bleporini:


sksamuel:
Great thanks!

elastic4s-core/src/test/scala/com/sksamuel/elastic4s/mappings/MappingDefinitionDslTest.scala
```diff
@@ -9,6 +9,26 @@ class MappingDefinitionDslTest extends WordSpec with Matchers with JsonSugar {
   import com.sksamuel.elastic4s.ElasticDsl._
 
   "mapping definition" should {
+    "insert source exclusion directives when set" in {
+      val mapping = new MappingDefinition("test").sourceExcludes("excludeMe1", "excludeMe2")
+      val output = mapping.build.string()
+      println("output = " + output)
```
martin-g: debug leftover

sksamuel: Thanks @martin-g , good spot. I've fixed it here baeb07e8eb77213b6a473c1668c693bad799116e

bleporini: My mistake ! Sorry for that ;-)

bleporini: Hello @sksamuel ,

Can you tell me when do you plan to publish a new release?

Regards!

PR #419, opened 2015-09-03
Adding return value to refresh() method

sjoerdmulder:
Handy if you want to `.await` the refresh action :)


sksamuel:
Good thanks.

PR #422, opened 2015-09-07
Prevent overrunning elasticsearch

sjoerdmulder:
When indexing is (very) busy but the interval is firing earlier than the BulkResponse returns this might cause the ForceIndexing command to execute while still processing. This will overrun elasticsearch (bulk.index.queue_size)


sksamuel:
Great thanks

PR #426, opened 2015-09-10
Replaced term and regex with termQuery and regexQuery

tototoshi:
`term` and `regex` are deprecated


sksamuel:
Good spot thanks. Also thank you for your CSV library, I use it a lot.

Issue #424, opened 2015-09-09
Indices Stats and Clear Indices with ambiguous implicit


luismfonseca:
I can't run the following code:
`esClient.execute(index stats(indexName))`

Error:
```
ambiguous implicit values:
[error]  both object IndicesStatsDefinitionExecutable in trait IndexAdminDsl of type com.sksamuel.elastic4s.ElasticDsl.IndicesStatsDefinitionExecutable.type
[error]  and object ClearIndicesCacheResponseExecutable in trait IndexAdminDsl of type com.sksamuel.elastic4s.ElasticDsl.ClearIndicesCacheResponseExecutable.type
[error]  match expected type com.sksamuel.elastic4s.Executable[com.sksamuel.elastic4s.admin.IndicesStatsDefinition,R,Q]
```

I think the line `com/sksamuel/elastic4s/admin/IndexAdminDsl.scala:56` is using `IndicesStatsDefinition` where it should be using something like `ClearIndiciesDefinition` to avoid this confusion.

If you believe this might be the fix the problem, I can make a pull request.

sksamuel:
Yes sounds right.

PR #427, opened 2015-09-10
Resolved ambiguous implicit issue #424

luismfonseca:
The `implicit object ClearIndicesCacheResponseExecutable` was incorrectly using `IndicesStatsDefinition`. This in turn caused ambiguousness when resolving the implicits.

(I've also changed the order of the classes declaration to match the order of the implicits)

This PR fixes issue #424.


sksamuel:
Thanks

PR #431, opened 2015-09-16
Fixing GET with slash notation: get id "id" from "index"/"type" now works.

henningp:
There was a bug when specifying the type as "index"/"type" (the type was ignored).


sksamuel:
Great spot thanks

henningp:
Spotted... in production, of course ;)

sksamuel:
If you use elastic4s in production, you could update the new "used by" section on the main readme if you wanted :)

PR #434, opened 2015-09-16
Sigterms add significanceheuristic

henningp:
Significant terms aggregation: Allowing to set the significance heuristic.


sksamuel:
Great, once agian, thanks.

PR #435, opened 2015-09-16
Geo_shape type: Allowing to specify "tree_levels" directly as an alternative to the "precision" option.

henningp:


elastic4s-core/src/main/scala/com/sksamuel/elastic4s/mappings/attributes.scala
```diff
@@ -508,6 +508,20 @@ object attributes {
     }
   }
 
+  trait AttributeTreeLevels extends Attribute { self: TypedFieldDefinition =>
+
+    private[this] var _treeLevels: Option[Integer] = None
+
+    def treeLevels(treeLevels: Integer): this.type = {
+      _treeLevels = Some(treeLevels)
```
martin-g: Since this is a `java.lang.Integer` it is possible to pass `null` so I think it is better to use `Option(treeLevels)` instead.

Issue #430, opened 2015-09-15
Validation of mapping parameters


andrestc:
Currently, I can define the following field mapping:

``` scala
"tags" typed CompletionType analyzer "standard"
```

Since [CompletionFieldDefinition](https://github.com/sksamuel/elastic4s/blob/952fe4dea4001bee700873b459357e35a5d15d6b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/mappings/fields.scala#L480) doesn't inherit from AttributeAnalyzer, that attribute will be ignored when generating the mapping.

I see two issues here (I can separate them into 2 different issues in github, but wanted to create this for discussion):
1. I believe we could do some kind of validation and warn the user when it tries to set an unsupported parameters. I can work on this, but would like to discuss the best approach.
2. I opened an [issue on elasticsearch](https://github.com/elastic/elasticsearch/issues/13584) because although its not documented, I was able to create an completion type using the "analyzer" parameter. So if it is indeed supported, we should add this to CompletionFieldDefinition. I can work on this once its confirmed.

andrestc:
As mentioned in the issue and stated in [2.0 docs](https://www.elastic.co/guide/en/elasticsearch/reference/2.0/search-suggesters-completion.html) "index_analyzer" is deprecated so we should deprecate these and use only "search_analyzer" and "analyzer".

sksamuel:
The deprecation can be added very quickly.

sksamuel:
Where does it say index_analyzer is deprecated btw ?

andrestc:
Yea, I [mentioned in my elasticsearch](https://github.com/elastic/elasticsearch/issues/13584#issuecomment-140496185) issue that they missed it in 1.7 docs. But in 2.0 docs you can see that [index_analyzers are not there anymore](https://www.elastic.co/guide/en/elasticsearch/reference/2.0/search-suggesters-completion.html). And  its not on [mapping parameters docs](https://www.elastic.co/guide/en/elasticsearch/reference/2.0/mapping-params.html) anymore.

sksamuel:
I see yes.

sksamuel:
Might be worth making a new issue to track that.

sksamuel:
I've updated 2.0 upgrade guide to explain this too as the deprecation will come too late for a lot of people.

andrestc:
Okay, i created an issue for the removal of ```indexAnalyzer``` for the 2.0 release.

[...]

PR #438, opened 2015-09-16
Deprecates indexAnalyzer

andrestc:
As discussed in #430 ```indexAnalyzer``` was deprecated and will be removed in 2.0.


sksamuel:
I liked the other one with 100 commits

andrestc:
:sweat_smile:

PR #441, opened 2015-09-17
Support for geo_shape filters.

henningp:


sksamuel:
This is great. Do you think you could get this into 2.0.0 on master too ?

henningp:
I'll give it a shot tomorrow.

project/Build.scala
```diff
@@ -35,8 +35,9 @@ object Build extends Build {
       "org.slf4j" % "log4j-over-slf4j" % Slf4jVersion % "test",
       "org.mockito" % "mockito-all" % MockitoVersion % "test",
       "org.scalatest" %% "scalatest" % ScalatestVersion % "test",
-      "org.codehaus.groovy" % "groovy" % GroovyVersion % "test"
-
+      "org.codehaus.groovy" % "groovy" % GroovyVersion % "test",
+      "org.codehaus.groovy" % "groovy" % GroovyVersion % "test",
```
martin-g: It seems the dependency for Groovy is doubled.

project/Build.scala
```diff
@@ -35,8 +35,9 @@ object Build extends Build {
...
+      "com.vividsolutions" % "jts" % "1.13" % "test"
```
martin-g: I don't see any new imports which should come from this new dependency.
Did I miss something?
Maybe it is an `optional` dependency in ElasticSearch ?!

henningp: Yes, exactly.

Issue #454, opened 2015-10-07
Histograms support for minDocCount


ddispaltro:
Can we add histograms support for min_doc_count? I see it's present for terms aggregation.

Thanks for the library!

MLnick:
@ddispaltro I took a quick look and this is a fairly simple PR - I have something just about ready. @sksamuel master is not compiling - is it broken? If so which branch should I PR against?

sksamuel:
Master is for 2.0 work. you can still put it there if u want or 1.7 branch

MLnick:
Ok will use 1.7



—

MLnick:
@ddispaltro can close this per #455

ddispaltro:
thanks

PR #455, opened 2015-10-12
Adding support for min_doc_count to Histgram aggregation (#454)

MLnick:
... also added other missing params (`extendedBounds` and `order`). Includes test case for default usage, `min_doc_count` and `order`.


ddispaltro:
@sksamuel this commit doesn't appear in the latest 2.1.0 or at least it's not in `HistogramAggregation`

sksamuel:
@ddispaltro looks like its just the test that was missing. I've added that as e7eb8bce5fa92410e2bdadce7fd1d6ecdcbccec8. Please confirm.

PR #471, opened 2015-11-27
fix: spell anaylzers -> analyzers

TimothyKlim:


sksamuel:
Oops, thanks for spotting that.

PR #482, opened 2015-11-30
Use Akka's Receive Timeout instead of the scheduler for forced bulk indexing

franbh:
The implementation of the forced bulk indexing causes useless scheduled operations during the normal operation of the BulkActor. This can be avoided using Akka's Receive Timeout feature (http://doc.akka.io/docs/akka/snapshot/scala/actors.html#Receive_timeout). Now forced bulk indexing is only performed when no messages arrive to the BulkActor in the specified timeout. Also, this allows to simplify the code (e.g. no need of BulkActor.ForceIndexing message anymore).

### DIFF
--- a/elastic4s-streams/src/main/scala/com/sksamuel/elastic4s/streams/BulkIndexingSubscriber.scala
+++ b/elastic4s-streams/src/main/scala/com/sksamuel/elastic4s/streams/BulkIndexingSubscriber.scala
@@ -1,6 +1,6 @@
 package com.sksamuel.elastic4s.streams
 
-import akka.actor.{Actor, ActorRef, ActorSystem, Cancellable, Props}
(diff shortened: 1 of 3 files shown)


sksamuel:
I like the idea, but what would happen if you set the timeout to 10 seconds, and a batch size of 100, and you get one message every 9 seconds. After 99 messages you've waited 99 x 9 seconds as each message resets the timeout right ?

franbh:
Yeah, you are right. It seems I've been too focused on our use case (massive bulk indexing).

franbh:
You can close the PR of course :-)

sksamuel:
I still think your idea is a good one, if people care about holding off the
batch if messages are still coming in. It's akin to Executors fixed delay
and Executors after period, in Java, if you know that code.

franbh:
Ok, I'll give it more thought.

PR #488, opened 2015-12-13
Support for include_in_root|parent

k4200:
This PR is to support `include_in_root` and `include_in_parent` even though they might be deprecated soon.
https://github.com/elastic/elasticsearch/issues/12461


sksamuel:
Great thanks. We can always add convenience methods later if they do deprecate them.

PR #489, opened 2015-12-15
Add parameter to ReindexDefinition to allow a QueryDefinition to be p…

Richard-Belcinski:
…assed.

This will allow the helper to be used in cases where re-indexing only a small subset of documents is required.  For instance, one might imagine "topping off" a new index with recently created items in a base index.


sksamuel:
Great thanks

PR #490, opened 2015-12-15
get request ignores type when using IndexType as param

mwielocha:
When using IndexType as a parameter for from method, index type is being ignored in Get requests.


sksamuel:
:+1:

PR #503, opened 2016-01-19
Add support for scripted agg metrics

nicoo:
https://www.elastic.co/guide/en/elasticsearch/reference/2.1/search-aggregations-metrics-scripted-metric-aggregation.html

- Change Script handling to not provide params when Map is empty.
This was causing:

org.elasticsearch.search.SearchParseException: init_script params are
not supported. Parameters for the init_script must be specified in the
params field on the scripted_metric aggregator not inside the
init_script object

- Update tests resources according to previous change

NB: I had a pull request merged in branch 1.7 for the same feature. Can you release version 1.7.5 as many of us are sticked to ES 1.7.x ?

Thanks


sksamuel:
Great, thanks.

PR #505, opened 2016-01-21
Type is required in put mapping request

alexshadow007:
Following code compiles:
```
client.execute {
      put mapping "index"
}
``` 
but fails with error:
```
head of empty list
java.util.NoSuchElementException: head of empty list
	at scala.collection.immutable.Nil$.head(List.scala:420)
	at scala.collection.immutable.Nil$.head(List.scala:417)
	at com.sksamuel.elastic4s.mappings.PutMappingDefinition.<init>(MappingDsl.scala:68)
...
```

This PR fixes this. Now you can't create put mapping request without type.


sksamuel:
:+1:

PR #512, opened 2016-02-03
Upgrate to elasticsearch 2.2.0

alexshadow007:
- MissingQuery is deprecated
- Field stats API now return values as number
- Fixing tests which use scripts. In 2.2.0 groovy plugin is separate artifact and local node doesn't load plugins from classpath. Using MockNode from Elasticsearch test framework.


sksamuel:
Brilliant thanks

PR #525, opened 2016-02-17
Query script file

Selsynn:
Adding a scriptType parameter in the scriptFieldDefinition, for usage in scriptfields.
Adding also a script for top hit aggregation.


sksamuel:
:+1:

PR #528, opened 2016-03-04
Add support for providing minScore in function score query

nosyrom:
Allows for specifying a "min_score" in a function score query without having to make calls directly on the builder instance.

### DIFF
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/SearchDsl.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/SearchDsl.scala
@@ -2,6 +2,7 @@ package com.sksamuel.elastic4s
 
 import java.util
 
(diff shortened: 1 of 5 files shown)


elastic4s-core/src/main/scala/com/sksamuel/elastic4s/SearchDsl.scala
```diff
@@ -128,7 +129,7 @@ case class RescoreDefinition(query: QueryDefinition) {
 }
 
 
-case class SearchDefinition(indexesTypes: IndexesAndTypes) {
+case class SearchDefinition(indexesTypes: IndexesAndTypes) extends DefinitionAttributeMinScore {
```
nosyrom: Was not sure if you eventually wanted to reuse the `DefinitionAttribute`s in the` SearchDefinition` as well.

sksamuel:
:+1:

PR #554, opened 2016-04-18
Added support for percolating the existing document by id.

eugengarkusha:
The operations priority :
1)id
2)rawDoc
3)fields


sksamuel:
👍

PR #555, opened 2016-04-27
Refactor Build.scala to AutoPlugin for future sbt 1.0

muuki88:
SBT 1.0 will remove Build.scala support.

This pull request refactors the Build into an AutoPlugin and upgrade the pgp plugin


sksamuel:
:+1:

PR #557, opened 2016-05-01
Proposal: HitAs codec derivation at compile time

muuki88:
This is a small draft on how code derivation for `elastic4s` could look like. Disclaimer: I'm a shapeless beginner and I am totally fine if this is too maintenance heavy and should be developed separately.

## The idea

Writing `HitAs` instances for each custom type is cumbersome ( same goes for `Indexable` ). So why not derive these type instances from case classes, so we can write:

```scala
import com.sksamuel.elastic4s.shapeless._
val resp = client.execute {
  search in "places" / "cities"
}.await
val cities: Array[City] = resp.as[City]
```

## Implementation

This compile time derivation can be done with shapeless. This initial draft has two versions and some limitations (which I would work on, if this has a chance of being merged)

### Limitations

- Nested objects can not be derived. The `sourceMap` is flat structure, thus the case classes need to be mapped to flat structure as well, building the correct map keys.
- No compile time verification that the defined `mappings` and `ADT` match
- No tests
- No docs

### The two versions:

The are two implementations `HitAsFromFieldsInstances` and `HitAsFromSourceInstances`.

- `HitAsFromFieldsInstances`
-- uses the `field` value to load values
-- needs fields specified in the query, e.g. `search in "places" / "cities" fields("id", "name")`
- `HitAsFromSourceInstances`
-- uses the `sourceMap` to parse values
-- needs no fields

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -9,7 +9,8 @@ lazy val root = Project("elastic4s", file("."))
     examples,
     jackson,
     json4s,
-    streams
+    streams,
+    shapeless
   )
 
 lazy val core = Project("elastic4s-core", file("elastic4s-core"))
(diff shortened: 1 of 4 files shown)


sksamuel:
Opted for #561

PR #561, opened 2016-05-05
Add circe-json based HitAs and Indexable instances

muuki88:
This is an addition or alternative to #557

[Circe](https://github.com/travisbrown/circe) provides automatic json codec
derivation, which makes it suitable for HitAs and Indexable derivations.

```scala
import io.circe.generic.auto._
import com.sksamuel.elastic4s.circe._

case class City(id: Int, name: String)

// index
index into "places" / "cities" id cityId source City(1, "munich")

// search and parse
val resp = client.execute {
  search in "places" / "cities"
}.await

val cities = resp.as[City]
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -52,6 +52,14 @@ lazy val jackson = Project("elastic4s-jackson", file("elastic4s-jackson"))
     libraryDependencies += "com.fasterxml.jackson.module" %% "jackson-module-scala" % JacksonVersion exclude("org.scala-lang", "scala-library"),
     libraryDependencies += "com.fasterxml.jackson.datatype" % "jackson-datatype-joda" % JacksonVersion
   ).dependsOn(core, testkit % "test")
+  
+lazy val circe = Project("elastic4s-circe", file("elastic4s-circe"))
+.settings(
+  name := "elastic4s-circe",
+  libraryDependencies += "io.circe" %% "circe-core" % CirceVersion,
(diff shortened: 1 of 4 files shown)


sksamuel:
I am reviewing your PRs. I'm just a bit busy atm, and they are quite detailed. Really good.

muuki88:
Thanks :) no pressure. I'll add docs as well, you approve this PR

muuki88:
If this gets accepted, we can add elastic4s to [this list](https://github.com/travisbrown/circe#related-projects) as well.

sksamuel:
👍

LATaylor-guardian:
@muuki88 @sksamuel is this yet to be released? struggling to find it on Maven.

sksamuel:
I've just released elastic4s 2.4.0 which has circe included although I couldn't get two of the tests to pass, so I ignored them. It looked like everything was good though @muuki88

muuki88:
Hi. Which tests failed? Strange that they pass on Travis 🤔

sksamuel:
It's the two that I commented out.
https://github.com/sksamuel/elastic4s/blob/master/elastic4s-circe/src/test/scala/com/sksamuel/elastic4s/circe/CodecDerivationTest.scala#L16

sksamuel:
It's probably something I broke when doing 2.4.0 but I can't seem to figure out what it is.

muuki88:
I'll take a look when I'm back home ( in a week or so)

Issue #562, opened 2016-05-06
ActorSystem in ScrollPublisher should be an ActorRefFactory


JD557:
The [ScrollPublisher](https://github.com/sksamuel/elastic4s/blob/master/elastic4s-streams/src/main/scala/com/sksamuel/elastic4s/streams/ScrollPublisher.scala) class requires an `implicit system: ActorSystem`, even though (AFAIK) it is only called to create a new actor on [line 44](https://github.com/sksamuel/elastic4s/blob/master/elastic4s-streams/src/main/scala/com/sksamuel/elastic4s/streams/ScrollPublisher.scala#L44)

I think this can and should be replaced with an [`ActorRefFactory`](http://doc.akka.io/api/akka/2.0/akka/actor/ActorRefFactory.html), so that an `ActorContext` could also be used.

I'm OK with sending a PR.

PR #563, opened 2016-05-06
Replace implicit ActorSystem with ActorRefFactory (in ReactiveElastic)

JD557:
This PR replaces the implicit `ActorSystem` required by `ReactiveElastic`with a `ActorRefFactory`.

This allows one to use either an `ActorSystem` or an `ActorContext`, which can be useful for some actor supervision strategies (i.e. the new actors can have a parent actor responsible for them).

Closes issue #562


sksamuel:
:+1:

Issue #478, opened 2015-11-29
Delete from index doesn't work using query.


tprizler:
I am using elastic4s version 2.0.0.
Looks like is no implementation to delete from an index.

Base on the following : https://github.com/sksamuel/elastic4s/blob/master/guide/delete.md the code should compile, but it doesn't:
         
    client.execute(ElasticDsl.delete from "myIndex" / "myType" where "email:my@gmail.com")

But only id works:

    client.execute(ElasticDsl.delete id "30" from myIndex" / "myType")

Make sense?

MLnick:
ES 2.0.0 removed delete by query- 


https://www.elastic.co/guide/en/elasticsearch/reference/1.6/docs-delete-by-query.html




—

sksamuel:
I need to update the docs to reflect that.

tprizler:
:+1: great!

PR #566, opened 2016-05-13
Remove reference to "delete from" in README

thomastoye:
delete from" has been deprecated and removed from elastic4s
Other references in the docs have been removed too, seems like this one was forgotten

See also: #478 & commit 375c46f


sksamuel:
:+1:

PR #569, opened 2016-05-17
Fixed race-condition when ensuring index existed

### LINKED ISSUE #552: ElasticSugar ensureIndexExists function race condition
This function checks if the index exists and then creates it if it doesn't. This is a race-condition: the index can be created in the time it takes between that check and the create index instruction.

I have some specs that do not work with this because for this reason (unless I force them to run sequentially).

This approach doesn't have this issue:
```
    client.execute {
      create index indexName
    }.recover {
      case _: IndexAlreadyExistsException => // Ok, ignore.
    }.await
```

I can make a MR if you agree with this changes :).

luismfonseca:
The previous approach was checking if the index existed and only then create the index if
necessary. This is a race-condition: the index can be created in the time it takes between
that check and the create index instruction.

Fixes #552.


sksamuel:
:+1:

PR #570, opened 2016-05-18
Add Ability to set SearchDefinition from json using extraSource or source

dgilling:
Port methods from Java API to scala SearchDefinition:
setExtraSource
setSource
setIndicesOptions


sksamuel:
:+1:

Issue #574, opened 2016-05-24
inner_hits options missing


jacool:
inner_hits options "_source" and "highlight" are missing

Issue #572, opened 2016-05-19
token_count missing options


jacool:
TokenCountDefinition should implement AttributeNullValue and AttributeStore

PR #575, opened 2016-05-24
inner_hits and token_count extensions

jacool:
Additions as per issues #574 and #572


sksamuel:
Sorry I seemed to have completed missed this!

PR #583, opened 2016-06-18
Option to classic null-check replacement in RichSearchHit class

kaching88:
When i worked with this library at work i noticed that in RichSearchHit class contains Option wrapper to null-check. Most of the times it's favourable to use Option but RichSearchHit is specific case. One response from elastic can contain many hits. When responses are small and three or ten additional  Some object inits aren't much deal but when we must parse something about one million hits those Options eats significant amount of time and memory. Of course i can do workaround by use .original but why RichSearchHit can't be more performance friendly?


sksamuel:
:+1:

PR #584, opened 2016-06-19
new method sourceAsMutableMap in RichSearchHit class

kaching88:
One of the most common use cases in work with ES is to get response, change it and send to frontend. Elastic4s in RichSearchResponse has only sourceAsMap which return immutable Map. Immutable maps are good choice for most cases, but often to meet above use case, we need mutable version to mutate values in efficient way. There isn't other convenient way to do this, cause most json libraries in scala, also works on immutable manner (f.e. play-json). For being precise, I'll do more often workaround by .original and converting with null-check to mutable.Map, than use immutable version from RichSearchResponse. The new method save me from writing these boilerplate repeatedly, when I need to use mutable.Map


sksamuel:
:+1:

PR #587, opened 2016-06-28
Return a failed future from the client when the java client raise an exception

fchaillou:
I'm starting to use your library which looks really great !

I just found one issue with the ElasticClient :

If we try to connect to a cluster that is down or misconfigured, the call to execute will throw an exception.
What i would expect to happen is to receive a failed future instead.

I would expect to be able to do that :

```scala
esClient
        .execute(get cluster health)
        .map(statusFromClusterHealth)
        .recover { case e : ElasticsearchException => e.failureNel }
```
but instead i have to do that :

```scala
try {
      esClient
        .execute(get cluster health)
    } catch {
      case e: ElasticsearchException => Future.successful(e.failureNel)
    }
```
The patch to this PR allow to get this behavior.
Thanks !
Fabien


sksamuel:
:+1:

PR #590, opened 2016-07-07
changed out sealed abstract classes for case classes as in 2.x

ericjohnston1989:
With the version 1.7.x there is no way to add a custom similarity because the similarity class is sealed and abstract. I've switched the attributes to case classes in the same way the 2.x branch is done.


sksamuel:
:+1:

PR #591, opened 2016-07-11
Previous pull request (converting sealed abstract to case) had compiler errors

ericjohnston1989:
Apologies! This one compiles.

case classes can't inherit case classes so I just swapped out the entire domain.scala file for the 2.x branch domain.scala file.


sksamuel:
:+1:

PR #593, opened 2016-07-12
added wrapperQuery

ericjohnston1989:
One of the QueryBuilders not implemented was the WrapperQuery. This allows us to include raw JSON for a specific query (for example, a new mapper definition defined in a script) while maintaining the DSL for the other parts of the search request.

btw -- is there a way to pull the latest snapshot from maven? Right now for the release/1.7.x branch I only pull 1.7.0 which was a fairly old commit. So currently I use elastic4s as an unmanaged dependency which is not ideal.


sksamuel:
awesome thanks

PR #596, opened 2016-07-16
Support for predefined language-specific stopwords

mkubala:
In order to customize a language-specific analyzer we need to reimplement it and add our custom filters/tokenizers/etc. (e.g.  https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-lang-analyzer.html#swedish-analyzer)

Unfortunately, it cannot be done using existing `StopTokenFilter`, which doesn't allow assigning string values to the `stopwords` field (only Arrays are supported).

This PR is aimed to solve this problem by introducing a `PredefinedStopTokenFilter` that accepts a plain string as a stopwords value.

Probably it could be done as well by changing `stopwords` type from `Iterable[String]` to some algebraic data type (`Either[String, Iterable[String]]` ?) but this way we will brake compability of the API.

Any kind of feedback will be welcomed!


sksamuel:
I will merge this in manually for 5.0

sksamuel:
Added manually something similar.

Issue #597, opened 2016-07-18
DSL using iterable parameters


jan0sch:
Hi,

I was trying to do something like this:

``` scala
val foo: Set[String] = ...
val bar: Set[String] = ...
val filterQueries: Set[QueryDefinition] = Set(termsQuery("foo", foo), termsQuery("bar", bar))
...
search(ElasticsearchIndexConfiguration.IndexName) query bool(
    must(matchQuery("file.content", queryString))
    should matchQuery("file.content.shingles", queryString)
    filter filterQueries
  )
...
```

This does not work because the `termsQuery` seems to ignore the iterable parameter. However the following workaround circumvents the issue:

``` scala
val filterQueries: Set[QueryDefinition] = Set(
  termsQuery("foo", foo.toArray:_*),
  termsQuery("bar", bar.toArray:_*)
)
```

sksamuel:
Dup of #599

PR #599, opened 2016-07-19
Support Iterables on the termsQuery DSL

JD557:
Fixes issue #597.

I needed to add some implicit objects and a new `BuildableTermsQuery` type class, but I'm not really fond of the name (`Buildable` sounds a bit weird for something that builds a builder).

I'm open to naming suggestions and other stylistic feedback (those implicit objects near the implicit definitions also look a bit out of place to me).


sksamuel:
Sorry I must admit to being a bit confused as to what you are doing here. Would you mind explaining it to an idiot like me.

JD557:
I started writing a quick explanation of what I was doing, only to notice that there's an easier way to do this (albeit "less typed").

I'll see if I can simplify this further, please do not review this yet.

sksamuel:
You might want to hold off anyway, I'm about to start upgrading the client to elasticsearch 5.0 and there's loads of breaking changes in 5.0, and so will be loads of breaking changes in e4s :)

sksamuel:
Inspired by this PR I've added support, basically the same as what you did.

PR #603, opened 2016-07-25
Restore deleted section in README.md

bigwheel:
"Using Elastic4s in your project" section was deleted in this commit https://github.com/sksamuel/elastic4s/commit/7220e1888a58c24f5e400591c61d55723d12fb7a .
However that is so useful, especially for elastic4s newbee like me.

If there is no reason, I want to restore it.


sksamuel:
Good point, thanks.

PR #607, opened 2016-08-01
Catch NoNodeAvailableException in ElasticClient

telser:
So this is a bit of an interesting issue. The exception above occurs sporadically on our system right now with the trace being that the Future produced by the client doesn't fail instead the exception happens.  It would appear from the modified code that the intent is to cleanly handle exceptions from the underlying elasticsearch.

Is this the correct place for this? If not, pointers on where to put this catch would be greatly appreciated.

Thanks,

Trevis


sksamuel:
We should probably just catch all exceptions. I can't remember why I limited it to those two.

PR #611, opened 2016-08-04
added highlighting, fetchSource, explain to TopHitsAggregator

ericjohnston1989:
TopHits aggregation allows for _explain and highlighting but this wasn't included in the API. I added these features. With highlighting I followed the highlighting method from the search api, so highlighting syntax is the same whether you are highlighting a search or a top hits aggregation.


sksamuel:
:+1:

PR #617, opened 2016-08-11
Add support for "missing" in all aggregations

luismcl:
Move the definition of missing method from GeoBoundsAggregationDefinition to ValuesSourceAggregationDefinition; then all sub-class of ValuesSourceAggregationDefinition has the missing method.


sksamuel:
:+1:

Issue #582, opened 2016-06-14
Can't use Pipeline Aggregation


romainfage:
Hi,

I want to use some Pipeline Aggregation, but I’m encountering an issue.

For example, I want to do this: 

``` json
{
    "aggs": {
        "foo1": {
            "terms": {
                "field": "field1",
                "order": {
                    "_term": "asc"
                },
                "size": 1
            },
            "aggs": {
                "foo2": {
                    "avg": {
                        "field": "field2"
                    }
                }
            }
        },
        "foo3": {
            "avg_bucket": {
                "buckets_path": "foo1>foo2"
            }
        }
    }
}
```

with this implementation

``` scala
aggs (
  termAggregation("foo1") field "field1" size 1 aggs {
                  avgAggregation("foo2") field "field2"
  },
  avgBucketAggregation("foo3") bucketPaths Seq("foo1", "foo2")
)
```

 but it didn’t compile :

``` bash
Type mismatch, found: AvgBucketDefinition, required: AbstractAggregationDefinition.
```

I looked into the source files and indeed `AvgBucketDefinition` extends `PipelineDefinition` which extends nothing.

Do I do something wrong with my implementation? Shouldn't `PipelineDefinition` extend `AbstractAggregationDefinition`, so that they can be used in an `aggs(…)` block as any other aggregation?

Thanks a lot

tarjeir:
Hit the same issue! Docs is not helping me out

PR #621, opened 2016-08-19
Pipeline aggregations do not extend abstract agg

tarjeir:
I encountered this issue earlier today : https://github.com/sksamuel/elastic4s/issues/582

I therefore suggest a change @romainfage mentioned in his post. Is it also natural to include the pipeline aggregations in the case object in the ElasticDsl trait? 

Regards
Tarjei


sksamuel:
Merged in the other duplicate pr.

PR #622, opened 2016-08-23
Fix wrong usage examples in guide/snapshot.md

k4200:
I'm not sure this is because the DSL might have changed in a recent version, but the examples in guide/snapshot.md don't compile with elastic4s 2.3.x.


sksamuel:
Thanks!

Issue #626, opened 2016-09-05
BulkIndexingSubscriber seems to hang after failure in bulk execution.


sjoerdmulder:
After receiving a java.lang.RuntimeException: Bulk operation failed too many times: failure in bulk execution in the logs it seems that the subscriber doesn't continue processing any other messages. Also the behaviour what happens is a bit unclear to me, could elaborate what happens to the actor when the RuntimeException is thrown? Shouldn't it be passed as message to self that the handleError() is called?

Also when there is this failure (in our case a StrictDynamicMappingException since the mapping is strict, and message invalid) how can we make sure that this message is skipped / dropped from the stream at one moment? E.g. keep retrying doesn't make sense in this case.

https://github.com/sksamuel/elastic4s/blob/master/elastic4s-streams/src/main/scala/com/sksamuel/elastic4s/streams/BulkIndexingSubscriber.scala#L205

sjoerdmulder:
Investigated somewhat more. When there is a Exception in the `onComplete` of a Future, it basicly checks if the error was NotFatal and then just logs the message.(See CallbackRunnable code in Promise.scala). 

So it will basicly just stop processing the stream and since its possible BulkActor.Result is not sent it will stall.

I am working on fixing this functionality, will submit PR

sjoerdmulder:
@sksamuel PR added, could you review and maybe publish / create 3.2.2 version :smiling_imp:

PR #627, opened 2016-09-06
Fixing BulkIndexSubscriber failure behaviour

sjoerdmulder:
Fix for issue #626

- Fixes the `hasSuccesses` functionality which was always false when there where failures but that doesn't have to be the case! Only a part of the BulkRequest can fail
- Instead of `RuntimeException` blowing it up `FailedResult` message is sent which will add to counter and call `onFailure` Response listener. This is also a change, now `onFailure` is only called when we give up. (e.g. after retries) instead of every retry, keeping it nice and clean for error handling.
- Added unittest to check for functionality (and that's why i found that the `hasSuccesses` was wrongly implemented)


sjoerdmulder:
#627 might also be resolved by this fix.

sksamuel:
Great thanks!

PR #637, opened 2016-09-27
Exposing missing methods from simple_query

jbjornson:
lenient, minimumShouldMatch and analyzeWildcard are all available on the builder but are not exposed in SimpleStringQueryDefinition


sksamuel:
Thanks I've merged this manually.

PR #645, opened 2016-10-13
QueryDSL improvements

danslapman:
- support `org.elasticsearch.common.unit.DistanceUnit.Distance` as argument for GeoDistanceQueryDefinition
- add geoDistanceRangeQuery to QueryDSL


sksamuel:
:+1:

PR #650, opened 2016-10-19
Introduce play-json support via elastic4s-play-json

danslapman:
This PR introduces Play-Json library support via elastic4s-play-json project


sksamuel:
:+1:

PR #653, opened 2016-10-30
Fix/derivation spec

muuki88:
Fixing the broken compilation test mentioned in #561


sksamuel:
:+1: What was the issue ?

muuki88:
The code snippet in the string could not be compiled , because some imports where missing.

I didn't investigate if it was a refactoring or automatic ide cleaning.

sksamuel:
Ok understood thanks. It was likely me removing them in intellij automatically.

PR #673, opened 2016-11-24
enable the X-Pack transport client 

gabriele83:


sksamuel:
Thanks for adding this. I've looked at the deps for x-pack and there's quite a lot so I'd be in favour of making this a seperate module 'elastic-xpack' which provides factory methods for making the client. That way, people not using it don't incur the extra deps. The other thing is, for some reason, elasticsearch people always use compile time scope in maven (wtf?) and so none of the transitives ever come down, so they would need to be added manually. I've traced the pom to here
https://artifacts.elastic.co/maven/org/elasticsearch/plugin/x-pack-api/5.0.0/x-pack-api.pom

sksamuel:
I'm happy to make the change if you want.

gabriele83:
Ok perfect.

Il 29 nov 2016 00:25, "Stephen Samuel" <notifications@github.com> ha
scritto:

PR #674, opened 2016-11-24
add TypedResponseListener to streams

rwiskerke:
If this PR is acceptable, I'll also supply a PR backporting this change to 2.3.x.

### DIFF
--- a/elastic4s-streams/src/main/scala/com/sksamuel/elastic4s/streams/BulkIndexingSubscriber.scala
+++ b/elastic4s-streams/src/main/scala/com/sksamuel/elastic4s/streams/BulkIndexingSubscriber.scala
@@ -25,7 +25,7 @@ import scala.util.{Failure, Success}
  */
 class BulkIndexingSubscriber[T] private[streams](client: ElasticClient,
                                                  builder: RequestBuilder[T],
-                                                 config: SubscriberConfig)
+                                                 config: TypedSubscriberConfig[T])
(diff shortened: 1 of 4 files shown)


elastic4s-streams/src/main/scala/com/sksamuel/elastic4s/streams/BulkIndexingSubscriber.scala
```diff
@@ -25,7 +25,7 @@ import scala.util.{Failure, Success}
  */
 class BulkIndexingSubscriber[T] private[streams](client: ElasticClient,
                                                  builder: RequestBuilder[T],
-                                                 config: SubscriberConfig)
+                                                 config: TypedSubscriberConfig[T])
```
sksamuel: This change means existing code will not compile, so I think we should add an apply method to do the conversion into a TypedSubscriberConfig[Nothing].

rwiskerke: This constructor is private to the `streams` package, so it shouldn't break any existing code of library users, right?

sksamuel:
Ah I missed that.

sksamuel:
Merged thanks.

rwiskerke:
If you create a 2.3 branch I will submit a PR for that as well. I already did the backport, but I don't think I can submit it without a branch which I can target?

rwiskerke:
I based my forks branch on commit 604cd06

PR #683, opened 2016-12-03
Fix link to IndexDslTest.

kurochenko:
Hi, I've found dead link in documentation and so i've created PR to fix it.


sksamuel:
:+1:

PR #694, opened 2016-12-14
Add support for collateQuery and collateParams

Dbzman:
I added support for collateQuery and collateParams for the phrase suggestions. This allows to filter out suggestions which don't yield any results.
Additionally I added 3 more tests for the phrase suggester which had none before.


sksamuel:
Awesome thanks.

PR #698, opened 2016-12-16
Prevent building of include/exclude when empty string is passed

Dbzman:
When using ` termsAggregation("agg1") field "job" includeExclude("something", "")` the exclude option would be built as well (and vice versa). I then realized that I have to pass "null" to not build that which I think should be abstracted away.

 This change was made to not build the include or exclude options when an empty string is passed. I did not change the parameters of the `def includeExclude(include: String, exclude: String): TermsAggregationDefinition` method to Option[String] because I did not find any other place in the DSL where this was a common practice.


sksamuel:
Sorry for delay on this :+1:

Dbzman:
Don't worry :) Thanks for merging!

Issue #717, opened 2017-01-30
Provide option to include exclude array of long / double values in termsaggregation definition


saucam:
Underlying java library provides the option to include exclude array of longs or doubles, so we should expose those options as well in, similar to array of strings:

```
TermsAggregationDefinition

def includeExclude(include: Iterable[String], exclude: Iterable[String]): TermsAggregationDefinition = {
    // empty array doesn't work, has to be null
    val inc = if (include.isEmpty) null else include.toArray
    val exc = if (exclude.isEmpty) null else exclude.toArray
    builder.includeExclude(new IncludeExclude(inc, exc))
    this
  }
```

PR #718, opened 2017-01-30
Provide option to include exclude array of long / double values apart from arrays of strings in TermsAggregationDefinition

saucam:
#717


sksamuel:
:+1:

solicode:
This doesn't appear to compile (due to type erasure + overloading).

sksamuel:
I've fixed the compile error.

PR #732, opened 2017-02-03
Fuzziness as a String in MatchQueryDefinition

### LINKED ISSUE #731: Revert fuzziness to String or use org.elasticsearch.common.unit.Fuzziness
In elastic4s 5.2.x (5.2.0 and 5.2.1) the match query API in elastic4s has changed to move fuzzyness to an Option[Double]. 
I was unable to pinpoint a specific commit which would explain this choice. The change was done in an overarching commit which changed a lot of things (https://github.com/sksamuel/elastic4s/commit/c0cbbb0e5f818ebdf5c527f2606b6829ed05b84b)  nor could I find any explanation in the elasticsearch java driver release notes or source code. 

This change effectively breaks `fuzziness("AUTO")` as there is no way to pass in "AUTO" anymore

`com.sksamuel.elastic4s.searches.queries.matches.MatchQueryBuilder` ends up calling `org.elasticsearch.index.query.MatchQueryBuilder#fuzziness` which delegates to  `org.elasticsearch.common.unit.Fuzziness#build` 

That factory method is implemented as follows 

```
    public static Fuzziness build(Object fuzziness) {
        if (fuzziness instanceof Fuzziness) {
            return (Fuzziness) fuzziness;
        }
        String string = fuzziness.toString();
        if (AUTO.asString().equalsIgnoreCase(string)) {
            return AUTO;
        }
        return new Fuzziness(string);
    }
```

So it either detects a `Fuzziness` instance or wraps a `String`. Passing it a double will get that double to be converted back to a String and wrapped. The "closest" there is to a double in the `Fuzziness` API is `org.elasticsearch.common.unit.Fuzziness#asFloat`
However the implementation is :
```
    public float asFloat() {
        if (this.equals(AUTO)) {
            return 1f;
        }
        return Float.parseFloat(fuzziness.toString());
    }

``` 
But that translation is unidirectional, try this in the console  :
```
import org.elasticsearch.common.unit.Fuzziness
Fuzziness.AUTO.equals(Fuzziness.build(Fuzzyness.AUTO.asFloat))
```
I get 
```
res1: Boolean = false
```

Unfortunately, that's what's used by `org.elasticsearch.common.unit.Fuzziness#asDistance(java.lang.String)` to actually determine the number of allowed edits ... 

```
    public int asDistance(String text) {
        if (this.equals(AUTO)) { //AUTO
            final int len = termLen(text);
            if (len <= 2) {
                return 0;
            } else if (len > 5) {
                return 2;
            } else {
                return 1;
            }
        }
        return Math.min(2, (int) asFloat());
    }
```

Since there is no way to build an instance of Fuzziness from a double which is equal to AUTO the AUTO fuzziness is effectively disabled by the new API.

jeantil:
fixes #731

### DIFF
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/searches/queries/matches/MatchQueryDefinition.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/searches/queries/matches/MatchQueryDefinition.scala
@@ -10,7 +10,7 @@ case class MatchQueryDefinition(field: String,
                                 analyzer: Option[String] = None,
                                 boost: Option[Double] = None,
                                 cutoffFrequency: Option[Double] = None,
-                                fuzziness: Option[Double] = None,
+                                fuzziness: Option[String] = None,
                                 fuzzyRewrite: Option[String] = None,
                                 fuzzyTranspositions: Option[Boolean] = None,
                                 lenient: Option[Boolean] = None,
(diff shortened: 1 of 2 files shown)


sksamuel:
Thanks

PR #740, opened 2017-02-09
Fix index create/delete with http client against a https server by in…

ppearcy:
…cluding leading forward slash on request endpoint

My test setup is running against a local node and AWS ES via HTTPS. 

Apparently there are some different semantics with leading / between HTTP vs HTTPS as the local node was fine while only HTTPS failed. 

Btw, thanks for the AWESOME library!


sksamuel:
Thanks.

PR #746, opened 2017-02-13
Add support for named filters in FiltersAggregation

To-om:
In Elastic4s 2.x, it was possible to provide name to [FiltersAggregation](https://github.com/sksamuel/elastic4s/blob/v2.4.0/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/aggregations.scala#L543) filters.

It seems that 5.x [FiltersAggregationDefinition](https://github.com/sksamuel/elastic4s/blob/master/elastic4s-tcp/src/main/scala/com/sksamuel/elastic4s/searches/aggs/FiltersAggregationDefinition.scala) accepts only [anonymous filters](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-filters-aggregation.html#_anonymous_filters).

This PR adds support to keyed FiltersAggregation.


sksamuel:
Thanks. We'll need a method in the agg dsl too so we don't have to use the case class name directly.

PR #752, opened 2017-02-15
Support for string parameters in CommonTermsQuery *minimumShouldMatch

### LINKED ISSUE #748: Support for non-integer types of minimum_should_match in CommonTermsQueryDefinition
In elastic search queries `minimum_should_match` parameter support different types of values. See https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-minimum-should-match.html
Same types are supported in `minimum_should_match` `low_freq` and `high_freq` parameters.
In older versions of elastic4s it was possible (corresponding methods accepted string parameter), but in current version only Integer type is supported.

tyth:
fixes #748


sksamuel:
Great thanks

PR #753, opened 2017-02-17
Fix for rawQuery

briantopping:
I noticed that `rawQuery` did not work due to a missing match, and it was unclear that it could work at all. With this code, a query of the form `{ "match": { "key": "value" } }` now works, see https://discuss.elastic.co/t/possible-to-pass-rest-json-query-format-to-the-java-api/2354 for additional information.


sksamuel:
:+1:

Issue #754, opened 2017-02-20
Optional dependencies causing stub warnings


solicode:
I noticed that the `Class ___ not found - continuing with a stub.` warning still exist in certain cases. For example, if you create a simple hello world application with just the following dependencies, the stub warnings will show up:

```scala
"com.sksamuel.elastic4s" %% "elastic4s-core" % "5.2.7",
"com.sksamuel.elastic4s" %% "elastic4s-http" % "5.2.7",
"com.sksamuel.elastic4s" %% "elastic4s-jackson" % "5.2.7"
```

When using `elastic4s-tcp`, the problem disappears.

I went looking for why that is, and it's because of how the dependencies are currently set up. Only the tcp project has the dependencies for `spatial4j`, `jts`, etc. `elastic4s-http` does not have these.

The real cause of all this is actually elasticsearch's use of optional dependencies [here](https://github.com/elastic/elasticsearch/blob/d235e489e6b36b4c84661b927481cd8cdcb87801/core/build.gradle#L79-L87).

To solve this, I only see 2 options:

1. Mark these dependencies Optional in all the elastic4s subprojects that use them. By doing this though, anybody who consumes elastic4s has to also add these dependencies in their own project (optionally marking them as `Optional` as well).
2. Simply make them compile dependencies in elastic4s (in core though, not just tcp). 

Personally, I like option 2 because optional dependencies seem to be quite problematic in Scala. It's up to you though. The tradeoff seems to be ease of use vs file size (due to potentially unnecessary jars being pulled in).

Since I already have option 2 ready, I'll create a PR. If you want to go with a different solution though, I can change it.

sksamuel:
Adding in the spatial4j is ok to HTTP as that's a small enough dep that I'm sure it won't bloat builds. As for the changes you made to TCP - if you remove the lucene deps, it'll break. They're needed to run the TCP client.

solicode:
Are they not already included with the `"org.elasticsearch.client" % "transport" % ElasticsearchVersion`? 

[transport](https://mvnrepository.com/artifact/org.elasticsearch.client/transport/5.2.0) -> [elasticsearch](https://mvnrepository.com/artifact/org.elasticsearch/elasticsearch/5.2.0)

Unless there's a reason to explicitly declare the transitive dependencies. I just figured it would cause more problems in the long run by accidentally pinning it to an older version of Lucene when it comes time to update `ElasticsearchVersion`.

sksamuel:
[...]

PR #755, opened 2017-02-20
Add optional dependencies to core

solicode:
Potential fix for #754

Also, I removed a bunch of dependency declarations that seemed unnecessary to me since they get pulled in transitively. If there's a reason for those though, let me know.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -18,13 +18,18 @@ lazy val root = Project("elastic4s", file("."))
 
 lazy val core = Project("elastic4s-core", file("elastic4s-core"))
   .settings(name := "elastic4s-core")
+  .settings(libraryDependencies ++= Seq(
+    "org.locationtech.spatial4j" % "spatial4j"     % "0.6",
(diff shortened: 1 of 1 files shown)


solicode:
I made a mess here with the commits. Let me get back to this later and clean it up.

solicode:
Alright, I cleaned up the history.

sksamuel:
There was some issue with log4j 2.7 which meant only 2.6.2 worked with elastic. Might have been fixed now mind.

sksamuel:
https://discuss.elastic.co/t/issue-with-elastic-search-5-0-0-noclassdeffounderror-org-apache-logging-log4j-logger/64262/4

solicode:
Ah, sorry, I didn't know about that. Haven't seen that problem myself. I was trying to match the version with the transitive dependency. Anyway, I reverted back to 2.6.2.

sksamuel:
Looks good now.

PR #758, opened 2017-02-20
Add support for range query in Http Client

Tyagi-Gaurav:
The support for Range Query was only available in TcpClient. This change is to make it available for HttpClient as well.


sksamuel:
Thanks.

PR #769, opened 2017-02-27
Support for include/exclude terms in SigTermsAggregationDefinition

tyth:


tyth:
It is already in the code, it's passed via `IncludeExclude` instance. 

https://github.com/sksamuel/elastic4s/blob/481e3f2fb608c6aeb4efc49384062716f9128e02/elastic4s-tcp/src/main/scala/com/sksamuel/elastic4s/searches/aggs/SigTermsAggregationBuilder.scala#L16

sksamuel:
Of course doh.

Issue #770, opened 2017-02-27
Missing SignificanceHeuristic in SigTermsAggregationDefinition


tyth:
Currently, there is no way to specify SignificanceHeuristic for `SigTermsAggregationDefinition`.
Even if a better scala-like API is planned for this type of parameter in the future, we need some workaround ("exposing" Java API via single method) now.

sksamuel:
Your PR will be included in 5.2.10

PR #772, opened 2017-02-28
SignificanceHeuristic parameter for SigTermsAggregation

tyth:
Potential (temporal?) fix for #770

Whether it is planned to hide Java `SignificanceHeuristic` class hierarchy from DSL users, it would be nice to be able to set that parameter right now.


sksamuel:
It's fine to use the Java one, as I'm not going to create Scala enums / case classes for things like this until its clear what the new Java REST client looks like (as it may very well provide these for us).

PR #785, opened 2017-03-07
Add include_lower, include_upper, and time_zone in range query

solicode:
This adds exclusive range support for `RangeQueryBodyFn`. I also noticed `time_zone` was missing, so I added that too.

It seems like `lt` and `gt` are preferred now, but since I wasn't 100% sure I didn't want to go renaming things. The Java client still uses `include_*`, so I guess it won't be removed anytime soon.


sksamuel:
:+1:

PR #787, opened 2017-03-07
#795 Add normalizer api

TrustNoOne:
I implemented the new normalizer API https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-normalizers.html

I don't know if you want to integrate it since it's still marked as experimental but I did it anyway.

The only non-backwards compatible difference is that calling .analysis(...) twice will now append the analyzers instead of overwriting them, I don't know if this could be a problem, feel free to take and/or modify the code as needed


sksamuel:
Thanks. I might update this later so that it uses immutable case classes which is the way I'm going with the rest of the API.

Issue #788, opened 2017-03-07
AliasExists endpoint is missing in the Http DSL 


mstockerl:
The [rest api](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-aliases.html) of Elasticsearch offers an endpoint to check, whether an alias exists.

PR #789, opened 2017-03-07
#788 add support for aliasExists check in Http DSL

mstockerl:
I noticed there are no (medium) tests for the HttpClient so far. 
Is this intended? If not I will start and write the first one for this endpoint.


sksamuel:
Hi. They are intended. It's because there's no HTTP support for the other alias operations yet. That's probably the most important missing http feature atm.

mstockerl:
Ok, good! We (gutefrage.net) currently moving from the Java Tcp Client to elastic4s Http Client due to incompatibilities with Spark ;) As we need these features too, we will contribute the other Alias operations soon.

sksamuel:
We use the http client in production at Deutsche Bank because of issues
with spark too.

PR #797, opened 2017-03-08
minor Circe improvements

yurique:
The `jsonFormatWithCirce` is currently doing `.right.get` which in case of problems with decoding just throws `NoSuchElementException("Either.right.value on Left")` which is not very informative. `fold(throw _, identity)` will throw the underlying Circe exception with information about what went wrong.

In the `indexableWithCirce` the parameter `printer: Json => String = Printer.noSpaces.pretty` allows to customize JSON encoding, like dropping `None` / `null` values, for example:
```
implicit val myJsonPrinter: Json => String = Printer.noSpaces.copy(dropNullKeys = true).pretty

esClient execute { 
  indexInto(index, documentType)
    .doc(doc)
    .id(doc.id)
}
```


sksamuel:
:+1:

Issue #743, opened 2017-02-10
SortDefinition can not be created. Missing addSort Method


alex-voigt:
```
java.lang.NoSuchMethodException: org.elasticsearch.search.builder.SearchSourceBuilder.addSort(org.elasticsearch.search.sort.SortBuilder)
[info]   at java.lang.Class.getMethod(Class.java:1786)
[info]   at com.sksamuel.elastic4s.search.SearchImplicits$SearchDefinitionShow$$anonfun$show$19.apply(SearchImplicits.scala:75)
[info]   at com.sksamuel.elastic4s.search.SearchImplicits$SearchDefinitionShow$$anonfun$show$19.apply(SearchImplicits.scala:74)
[info]   at scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:33)
[info]   at scala.collection.mutable.WrappedArray.foreach(WrappedArray.scala:35)
[info]   at com.sksamuel.elastic4s.search.SearchImplicits$SearchDefinitionShow$.show(SearchImplicits.scala:74)
[info]   at com.searchperience.frontend.core.builder.HighlightDefinitionBuilderSpec$$anonfun$1.apply$mcV$sp(HighlightDefinitionBuilderSpec.scala:45)
[info]   at com.searchperience.frontend.core.builder.HighlightDefinitionBuilderSpec$$anonfun$1.apply(HighlightDefinitionBuilderSpec.scala:12)
[info]   at com.searchperience.frontend.core.builder.HighlightDefinitionBuilderSpec$$anonfun$1.apply(HighlightDefinitionBuilderSpec.scala:12)
[info]   at org.scalatest.Transformer$$anonfun$apply$1.apply$mcV$sp(Transformer.scala:22)
```

Seems like the addSort Method is not existing in the SortBuilder.java in Elasticsearch when called by elastic4s 5.2.5 SearchBuilderFn (line 53) and SearchImplicits (line 75).

sksamuel:
Seems like a version mismatch. Do you have earlier versions of e4s lurking on your classpath.

Dbzman:
We looked into the elastic4s source code and noticed that you have the following lines in `SearchBuilderFn`:
```
  if (search.sorts.nonEmpty)
      search.sorts.foreach { sort =>
        builder.getClass.getMethod("addSort", classOf[SortBuilder[_]]).invoke(builder, SortBuilderFn.apply(sort))
      }
```
[...]

PR #800, opened 2017-03-08
fixed sort issue on 5.2

alenegro81:
I should have fixed this bug: https://github.com/sksamuel/elastic4s/issues/743
I tested on my own software that present that issue and now it is solved.


sksamuel:
Thanks.

alenegro81:
could you make a new release?

Issue #565, opened 2016-05-10
Create index from raw JSON


Iulian7:
This is also on stackoverflow: http://stackoverflow.com/questions/37138949/create-index-from-raw-json-using-elastic4s

Basically I need to create an index with completion context mappings as in (https://github.com/sksamuel/elastic4s/issues/452). As I understand this is not yet supported by the DSL.

Is it possible to create an index from a raw JSON string?

Thanks, 
Iulian

brstos:
Considering that you have the JSON mapping in a variable `rawMapping` like this:

    val rawMapping =
        """{
              "service": {
                      "properties": {
                          "name": {
                              "type" : "string"
                          },
                          "tag": {
                              "type" : "string"
                          },
                          "suggest_field": {
                              "type": "completion",
                              "context": {
                                  "color": {
                                      "type": "category",
                                      "path": "color_field",
                                      "default": ["red", "green", "blue"]
                                  },
                                  "location": {
                                      "type": "geo",
                                      "precision": "5m",
                                      "neighbors": true,
                                      "default": "u33"
                                  }
                          }
                      }
                  }
                  }
              }"""

You can create the index using the raw mapping like this:

    client.execute {
        create index "services" source rawMapping
    }

salvob41:
This is a really interesting feature. Thank you! 
I just want to update the answer by @brstos . From 5.6 the API wants the rawMapping with the `"mappings"` key 

Updated example: 
```
 val rawMapping =
    """{
      "mappings":{
          "service": {
                  "properties": {
                      "name": {
                          "type" : "string"
                      },
                      "tag": {
                          "type" : "string"
                      },
                      "suggest_field": {
                          "type": "completion",
                          "context": {
                              "color": {
                                  "type": "category",
                                  "path": "color_field",
                                  "default": ["red", "green", "blue"]
                              },
                              "location": {
                                  "type": "geo",
                                  "precision": "5m",
                                  "neighbors": true,
                                  "default": "u33"
                              }
                      }
                  }
              }
              }
          }
   }"""
```

and so the API to call is: 
```
client.execute {
    createIndex("services") source rawMapping
}
```

PR #817, opened 2017-03-13
#565 create index from raw json string also with Http Client

mstockerl:
The suggested solution in this ticket does not work with the Http Client


mstockerl:
Also added SubAggregation support in TermsAggregation

sksamuel:
Can terms have sub aggregations?

mstockerl:
Yes, it is a bucket aggregation. 
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-terms-aggregation.html

sksamuel:
Ah yeah of course. Should probably do the metadata fields too.

PR #831, opened 2017-03-17
Add implicit codec when converting a http response to a string

muuki88:
## Goal

Define the `scala.io.Codec` ( string encoding ) that should be used when converting
a http response to a string.

## Fixes

When your `defaultCharset` defined in `scala.io.Codec` doesn't match the http response encoding.


sksamuel:
Is there a default Codec in scope?

muuki88:
Yes. The `Source.fromInputStream(r.getEntity.getContent)` already uses this and it defaults to the `java.nio.Charset.defaultCharset`.

The `scala.io.codec` companion object defines this default codec and is thus being automatically in scope.

PR #835, opened 2017-03-18
DateRangeAggregation support for multiple unbounded ranges

tyth:


sksamuel:
We should probably add methods that accept an iterable too.

PR #836, opened 2017-03-18
Enable parallel tests

solicode:
The tests look properly isolated to me with how embedded nodes are managed, so I was curious what would happen if I enable parallel tests. All the tests still pass and I'm seeing a *very* significant speedup (close to 4x) on my machine for running all the tests: 453s (serial) vs 116s (parallel)

Running tests in parallel can sometimes complicate things, so whether you want to enable them is completely up to you. The work required to set them up seem to already be complete though from what I can tell.

I'm generally in favor for parallel tests because besides the speedup, they can often expose bugs.

I'm curious how the tests will behave on Travis. I wouldn't expect a big speedup there like you would on a dev machine, but let's see what happens. At the very least, I want to confirm that nothing breaks due to this.

By the way, for this I also re-enabled `elastic4s-tests` so that it runs with `sbt test`.


solicode:
Hmm, all green on Travis, and it took 11 mins to complete compared to the previous ~17 mins, and that's not including the fact that `elastic4s-tests` weren't running before this. That's actually quite surprising to me.

sksamuel:
This is great.

PR #840, opened 2017-03-21
make RichSearchHit#fieldValueOpt return an Option[AnyRef]

yurique:
make `RichSearchHit#fieldValueOpt` return an `Option[AnyRef]` (was just `AnyRef`)


sksamuel:
Nice spot thanks.

PR #852, opened 2017-03-30
DateHistogramAggregation support for HttpClient

rkrzewski:
I need to use date_histogram_aggregation in a query but it turned out that a builder object for this particular aggregation was not implemented yet. Luckily it was quite easy to implement taking existing builders as example.


sksamuel:
:+1:

PR #854, opened 2017-04-02
Wrongly named POST parameter

cchantep:
versionType/version_type: 
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update.html


sksamuel:
:+1:

PR #856, opened 2017-04-04
Adding PutMapping implementation in Http client

alex1712:
As we needed to use the PutMapping in Http in 1 of our applications I added the implementation in the library, I hope I didn't forget anything.

Thank you very much for your work :)


elastic4s-core/src/main/scala/com/sksamuel/elastic4s/mappings/MappingDefinition.scala
```diff
@@ -41,13 +45,15 @@ case class PutMappingDefinition(indexesAndType: IndexesAndType,
                                 routing: Option[RoutingDefinition] = None,
                                 timestamp: Option[TimestampDefinition] = None,
                                 templates: Seq[DynamicTemplateDefinition] = Nil,
-                                id: Option[IdField] = None
+                                id: Option[IdField] = None,
```
sksamuel: I think this can be removed as you can't set anything on id field anymore.

alex1712: I tried it in the test to check if the request get rejected or something and it seems that this parameter won't be reflected in the Put body at all. So I will remove it.

Thank you for the comment :)

sksamuel:
awesome thanks

Issue #824, opened 2017-03-14
Http Client - GetAlias request throws class cast exception on missing index


muuki88:
Hi,

When I run a `GetAlias` request and there are no aliases the client throws an exception instead of handling the empty response.

### expected behaviour

Either a failed Future, but I prefer a valid response with an empty Map.

### steps to reproduce

```scala
import com.sksamuel.elastic4s.http.HttpClient
import com.sksamuel.elastic4s.http.ElasticDsl
import com.sksamuel.elastic4s.http.ElasticDsl._
import scala.concurrent.duration.Duration

val httpClient: HttpClient = ???
val alias: Seq[String] = Seq("non-existing-alias")

// This call throws a parsing exception
val getAliasResponse = Await.result(httpClient.execute {
    getAlias(alias)
}, atMost = Duration(5, TimeUnit.SECONDS))

getAliasResponse.map {
    case (indexName, aliasesMap) =>
       val aliases = aliasesMap.getOrElse("aliases", Map.empty).keys
}.toSeq
```

If you call the rest endpoint directly with

```
_alias/non-existing
```

The response is

```json
{
  "error": "alias [non-existing-alias] missing",
  "status": 404
}
```

which the `GetAliasExecutable` is unable to parse, because the format is

```cala
type GetAliasResponse = Map[String, Map[String, Map[String, AnyRef]]]
```

solicode:
I haven't looked too much into it, but this likely has to do with [this section](https://github.com/sksamuel/elastic4s/blob/4b80b9fa941d55d9e1bf067bcee2f6b25dfa8a36/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/HttpClient.scala#L86-L94) I added fairly recently.

I made that change because there were many instances where exceptions were being thrown, when really the content itself was a "valid" response (for example, GetResponse where the `found: Boolean` flag on it is false). You'd never get that GetResponse because of the exception being thrown.

But as you can see, the implementation isn't complete. It fixed a few endpoints, but broke this one.

The way Elastic implemented the rest client makes handling this a bit difficult because I don't see a simple way of differentiating between these errors and "true" exceptions. I'm still not sure what the best way of handling this is. I wanted to avoid trying to parse twice, but I'm not sure what other options there are. I was also trying to avoid changing JsonFormat.

I can look into it more, but if anybody has some other ideas, I would appreciate it.

sksamuel:
[...]

PR #868, opened 2017-04-14
Feature/httpjson

sksamuel:


sksamuel:
@solicode @muuki88 Have a look if you don't mind :)

muuki88:
Nice. Will take a look when I'm back from Dublin 😊

sksamuel:
Don't drink too much Guiness :)

muuki88:
Your avatar image makes this advice hard to take seriously 🤣😆

sksamuel:
Ha, funny that was in Germany actually. Can't remember where though.

sksamuel:
Are you sober yet @muuki88  :)

muuki88:
Yeah 🤣 Very busy week. I only had a brief chance to look over your changes. 

On the fist view it looks good. However such a big change deserves more attention and detailed feedback. Sorry for keeping you waiting, but I want to review this with the necessary attention 😃

solicode:
Sorry, I didn't notice this until now.

I don't really have much to add. The changes look fine to me (and the code is a lot cleaner too). It certainly makes consuming elastic4s less confusing for users, especially for those using different JSON libraries (like play-json) for `Indexable` and `HitReader`.

For me, the biggest pain point has been the aggregations. I know it was already discussed in the other thread, but working with `Map[String, AnyRef]` is quite difficult. I find myself wanting to work with the aggregation JSON directly and just leverage whatever JSON library I'm using to map it to proper types. Since aggregations can be nested, if I try to work with `Map[String, AnyRef]`, I get tempted to create some sort of mapping/decoding construct that's recursive... which somewhat resembles what a JSON library does using typeclasses. It seems like something that could be solved at the elastic4s level, but I still can't think of a solution that I like (and one that avoids double-parsing).

Other than that, I can't really think of anything else. @muuki88 probably has more to add than me regarding this topic.

elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/ResponseHandler.scala
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// a ResponseHandler that marshalls the body into the required type using Jackson
+// the response body is converted into a string using a codec derived from the content encoding header
+// if the content encoding header is null, then UTF-8 is assumed
+object ResponseHandler {
+
+  def fromEntity[U: Manifest](entity: HttpEntity): Try[U] = {
+    Try {
+      val charset = Option(entity.getContentEncoding).map(_.getValue).getOrElse("UTF-8")
```
muuki88: :+1: for handling the encoding :)

elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/HttpExecutable.scala
```diff
@@ -0,0 +1,49 @@
...
+    private def future(callback: ResponseListener => Unit, handler: ResponseHandler[U]): Future[U] = {
+      val p = Promise[U]()
+      callback(new ResponseListener {
+        override def onSuccess(r: Response): Unit = p.tryComplete(handler.onResponse(r))
+        override def onFailure(e: Exception): Unit = p.tryComplete(handler.onError(e))
+      })
+      p.future
+    }
+
+    def future(method: String,
```
muuki88: Why did you rename this from `performAsyncResponse` to `future`? This method name is a bit confusing. Some API docs would  be nice on this one.

elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/ResponseHandler.scala
```diff
@@ -0,0 +1,67 @@
...
+
+trait ResponseHandler[U] {
+  def onResponse(response: Response): Try[U]
+  def onError(e: Exception): Try[U]
+}
+
+// a ResponseHandler that marshalls the body into the required type using Jackson
+// the response body is converted into a string using a codec derived from the content encoding header
+// if the content encoding header is null, then UTF-8 is assumed
+object ResponseHandler {
```
muuki88: I guess this scaladoc belongs to the `ResponseHandler` trait and should  be written as scaladoc.

Even this is used internally, it should be more documented on _why_ it exists rather then implementation details like "utf-8" is used if missing. If I understand correctly from the other changes a `ResponseHandler` defines the behaviour regarding the http status codes. 

This trait is very similiar to [twitter finagle's response classifier](https://twitter.github.io/finagle/guide/Clients.html#response-classification). However the response classification is _only_ for classification, not for parsing, which I think is a good decision. The type definition is

```scala
type ResponseClassifier = ParticalFunction[ReqRep, ResponseClass]
```

In our case this could be something like

```scala
type ResponseClassifier = ParticalFunction[ResponseException, Try[U]]
```

However using the exception directly isn't a very good api design. Ideally the http response itself can be used. Not sure if `204` No Content can be an error as well. And what about `5xx` codes? When you use `failure404` does this exclude `5xx` codes? I'm asking, because this maybe confusing for other contributors at first sight.

Not sure if this is a good fit for the apache http implementation if it works with exception for propagating non `2xx` status codes. Still it works better for scala devs I guess, if you have meaningful types to work with,

elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/ResponseHandler.scala
```diff
@@ -0,0 +1,67 @@
...
+  def default[U: Manifest] = new DefaultResponseHandler[U]
+  def failure404[U: Manifest] = new NotFound404ResponseHandler[U]
+}
+
+class DefaultResponseHandler[U: Manifest] extends ResponseHandler[U] {
+  override def onResponse(response: Response): Try[U] = ResponseHandler.fromEntity[U](response.getEntity)
+  override def onError(e: Exception): Try[U] = Failure(e)
+}
+
+class NotFound404ResponseHandler[U: Manifest] extends DefaultResponseHandler[U] {
```
muuki88: If you extend the `DefaultResponseHandler` then `onResponse` can be removed and you should call `super.onResponse` for the "default" case

elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/index/IndexTemplateImplicits.scala
```diff
@@ -71,7 +65,7 @@ object CreateIndexTemplateBodyFn {
     if (create.mappings.nonEmpty) {
       builder.startObject("mappings")
       create.mappings.foreach { mapping =>
-        builder rawValue MappingContentBuilder.buildWithName(mapping, mapping.`type`).bytes()
+        builder.rawValue(MappingContentBuilder.buildWithName(mapping, mapping.`type`).bytes, XContentType.JSON)
```
muuki88: :+1:  for using `.` syntax and braces

elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/index/alias/IndexAliasImplicits.scala
```diff
@@ -1,47 +1,46 @@
...
-                         format: JsonFormat[GetAliasResponse]): Future[GetAliasResponse] = {
+                         request: GetAliasDefinition): Future[GetAliasResponse] = {
       val indexPathElement = if (request.indices.isEmpty) "" else s"/${request.indices.mkString(",")}"
       val endpoint = s"$indexPathElement/_alias/${request.aliases.mkString(",")}"
       logger.debug(s"Connecting to $endpoint for get alias request")
 
-      executeAsyncAndMapResponse(client.performRequestAsync("GET", endpoint, _), format, {
-        case re: ResponseException if re.getResponse.getStatusLine.getStatusCode == 404 => Try(Map.empty)
-        case e => Failure(e)
+      client.future("GET", endpoint, Map.empty, new ResponseHandler[GetAliasResponse] {
```
muuki88: Why not use the `failure404` handler?

solicode: Because of #824. 

I didn't have a strong preference either way. I just went with `Map.empty` because you said "I prefer a valid response with an empty Map." The other option is of course going with a failed Future.

solicode: Ah, I think I misunderstood. I assumed `failure404` was doing the opposite of what it actually does for some reason.

solicode: Still though, `failure404` wouldn't work because you just get JSON with the contents of the error. It doesn't map to the "expected" response entity the same way `GetResponse` and so on do.

elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/HttpExecutable.scala
sksamuel: Can add docs of course. It's an internal method.

sksamuel:
@solicode we could add an `.aggregationsAs[T]` which allows you to get a strongly typed object from the aggregations json. Then if you don't want to do that you can continue to use `termAggs(name)` or whatever as is current.

elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/ResponseHandler.scala
sksamuel: The idea behind `ResponseHandler` is that it handles the complete response, not just exceptions. It means that requests that need to handle 404s or other things can be supported by just writing another handler. So if we have a request type that returns text, or a request type that handles 204s specially, or whatever, it can be done easily. 

So its like the original error handler but expanded. 

Of course most of the elastic responses are just "marshall to an object" and so that's what the `DefaultResponseHandler` does.

sksamuel:
Might as well merge this then?

solicode:
No objections here.

Also, the `.aggregationsAs[T]` solution sounds reasonable to me, at least for what I'm trying to accomplish.

PR #869, opened 2017-04-15
Snapshot test suggestion

grillorafael:
Hey @sksamuel.

Just a test change suggestion. Would like to know what do you think.

Start using snapshot matcher in favor of match json resource and simple json string match.

I put a commend on the test that actually is using this matcher.

You can find more info here: https://github.com/commodityvectors/scalatest-snapshot-matchers

### DIFF
--- a/elastic4s-core/src/test/scala/com/sksamuel/elastic4s/mappings/MappingDefinitionDslTest.scala
+++ b/elastic4s-core/src/test/scala/com/sksamuel/elastic4s/mappings/MappingDefinitionDslTest.scala
@@ -1,61 +1,68 @@
 package com.sksamuel.elastic4s.mappings
 
+import java.io.{StringWriter, Writer}
+
+import com.commodityvectors.snapshotmatchers.SnapshotMatcher
+import com.fasterxml.jackson.core.util.DefaultPrettyPrinter
+import com.fasterxml.jackson.databind.{ObjectMapper, SerializationFeature}
+import com.fasterxml.jackson.module.scala.experimental.ScalaObjectMapper
 import com.sksamuel.elastic4s.{ElasticApi, JsonSugar}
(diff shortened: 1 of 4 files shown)


elastic4s-core/src/test/scala/com/sksamuel/elastic4s/mappings/MappingDefinitionDslTest.scala
```diff
@@ -64,31 +71,32 @@ class MappingDefinitionDslTest extends WordSpec with Matchers with JsonSugar wit
...
-    "include timestamp store field" in {
+    "include timestamp store field" in { implicit test =>
       val req = createIndex("docsAndTags").mappings(
         mapping("foo") timestamp {
           timestamp(true) path "mypath" store true
         }
       )
-      CreateIndexContentBuilder(req).string() should matchJsonResource("/json/mappings/timestamp_store.json")
+
+      CreateIndexContentBuilder(req) should matchSnapshot[XContentBuilder]()
```
grillorafael: This one

sksamuel:
Is this your project? If so, I'd consider adding some explanation to the readme on what it does. I've read the front page and I'm not sure what a snapshot is and why it would be useful to use it.

grillorafael:
Yes it's mine. It's a scalatest implementation of Jest's snapshot testing (https://www.youtube.com/watch?v=HAuXJVI_bUs). Basically what it does is it facilitates matching by automatically generating your match file and when there is a mismatch, showing the diff and prompting for updates.

You can run the change test changing the json content to see how it works as well. It might be helpful for testing json like strings like this. We are currently using in my current job's projects and it is working quite well for us.

That is just a suggestion in case you think there is a fit for elastic4s.

I'll try to improve the readme as well

grillorafael:
I'll close this as it isn't hugely important and will not bring much improvements overall

PR #875, opened 2017-04-19
Fixes 'type' and 'matched_fields' for highlight query.

jkugiya:
This PR will fix query for highlighting.

Problems are 
- `type` is ignored.
- Multiple `matched_fields` are generated.(such like `{ "matched_fields": "text", "matched_fields": "text.japanese" }`)


sksamuel:
:+1:

PR #876, opened 2017-04-19
Supports inner highlight

jkugiya:
This PR enables to supports inner hit's highlight.(maybe partially).

### DIFF
--- a/elastic4s-http/src/test/scala/com/sksamuel/elastic4s/http/search/queries/nested/NestedQueryBodyFnTest.scala
+++ b/elastic4s-http/src/test/scala/com/sksamuel/elastic4s/http/search/queries/nested/NestedQueryBodyFnTest.scala
@@ -1,0 +1,33 @@
+package com.sksamuel.elastic4s.http.search.queries.nested
+
+import com.sksamuel.elastic4s.http.ElasticDsl._
+import org.scalatest.{FunSuite, Matchers}
+
+class NestedQueryBodyFnTest extends FunSuite with Matchers {
+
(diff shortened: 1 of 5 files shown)


elastic4s-http/src/test/scala/com/sksamuel/elastic4s/http/search/queries/nested/NestedQueryBodyFnTest.scala
```diff
@@ -0,0 +1,33 @@
...
+        highlight("messages.text")
+          .matchedFields("messages.text", "messages.japanese")
+      }
+    }
+    NestedQueryBodyFn(query).string() shouldBe
+      "{\"nested\":{\"path\":\"messages\",\"query\":{\"match_all\":{}},\"inner_hits\":{\"highlight\":{\"fields\":{\"messages.text\":{\"matched_fields\":[\"messages.text\",\"messages.japanese\"]}}}}}}"
+  }
+  test("inner highlight with 'highlighterType' generates proper 'type' field.") {
+    val query = nestedQuery("messages", matchAllQuery()) inner {
+      innerHits("") highlighting {
```
jkugiya:
```java
   // InnerHitBuilder.java
   @Override
    public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
        builder.startObject();
        if (name != null) {
            builder.field(NAME_FIELD.getPreferredName(), name);
        }
```
It seems `InnerHitBuilder` expects name is optional, isn't it?

sksamuel:
Awesome.

PR #877, opened 2017-04-19
Various http bugs

zarucki:
Couple bugs I found when using http client in some of tcp tests
- Weird one with separator missing when constructing array with rawValue. I issued bug on elastic [repo](https://github.com/elastic/elasticsearch/issues/24166)
- Just one extra endObject too much
- Little bit different json for common query

Sorry for lack of tests.


sksamuel:
Awesome thanks

PR #879, opened 2017-04-20
Add support of range aggregation for http client

FabienPennequin:
This is my little contribution… to your great Elasticsearch client for scala ;-)

I added support of the range aggregation for http client.

I will try to add the others range aggregations (ip and date) in anothers PRs


sksamuel:
Awesome.

PR #882, opened 2017-04-22
Makes highlights more DRY

jkugiya:
- Makes creating JSON for highlighting more DRY
- Fixes `boundary_chars`'s value creation. (That had created string such like `C@2f7d905a`)
- Adds more tests.


sksamuel:
:+1:

PR #888, opened 2017-04-24
Exposed format option for RangeQueryDefinition

ryan-martin-axiell:
Added a "format" option for range queries. Based on the [Range Query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-range-query.html#_date_format_in_range_queries).


sksamuel:
:+1:

PR #891, opened 2017-04-25
fix StackOverflowError for GeoBoundingBoxQueryDefinition.corners

maksim07:
The method called itself which caused StackOverflowError


sksamuel:
:+1:

PR #898, opened 2017-05-03
Add simple category contexts for completion suggestions

Dbzman:
This is a basic implementation for getting context suggestions to work.
It currently covers our use case.


sksamuel:
I've merged this but tweaked it a bit to include boosts and prefixes on the contexts. I hope that's ok.

Dbzman:
Sure, that's perfect. 👍
Thank you very much!

PR #918, opened 2017-05-23
handling FunctionScoreQuery, sum score_mode added, track_scores flag …

adrian83:
…handled, sorting by _score and GeoDistance


sksamuel:
Awesome thanks. If you want this to be released as part of 5.4 then you'll need to do a PR against that branch too.

mbarroux:
Hello, I would need these improvements for 5.4.x if possible to PR against that branch please ;)

Thank you in advance,

PR #934, opened 2017-06-02
Clean up build.sbt

bbroeksema:
Most lucene requirements are not needed and joda-time is actually
required for compiling core. This was actually required to make the
project build from scratch.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -30,8 +30,10 @@ lazy val root = Project("elastic4s", file("."))
 lazy val core = Project("elastic4s-core", file("elastic4s-core"))
   .settings(name := "elastic4s-core")
   .settings(libraryDependencies ++= Seq(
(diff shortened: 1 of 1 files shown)


sksamuel:
The lucene requirements are needed for gradle builds.

bbroeksema:
right. could you tell me how i can test that? I'm using sbt mainly. I'm fine with further improving the patch as long as I know how to test.

I guess we aggree that core is the proper place for the requirements?

sksamuel:
Well you'd need to setup a gradle project, and include it. You can't remove them because like I said, the transitivity doesn't work for gradle for some reason. I'm happy if you know a workaround.

However, what I'd really recommend is you upgrade to the HTTP client because in v6 onwards, there's no dependency on elasticsearch core at all.

bbroeksema:
re workaround: stop using gradle?! ;). Anyway, master is building for me, I'll close this pr.

PR #936, opened 2017-06-05
More-like-this query routing fix

tyth:
`more_like_this` query does not work for parent-child relationship without explicit routing parameter, and there was no way to specify it. 

I changed `MoreLikeThisItem` a bit , but it should support old API calls.

ElasticSearch team recently fixed a linked bug https://github.com/elastic/elasticsearch/issues/23699 (routing parameter in `more_like_this` was ignored), so Elastic version had to be updated to 5.4.1

### DIFF
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/searches/QueryApi.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/searches/QueryApi.scala
@@ -161,13 +161,13 @@ trait QueryApi {
       likeItems(first +: rest)
 
     def likeItems(items: Iterable[MoreLikeThisItem]): MoreLikeThisQueryDefinition =
-      likeDocs(items.map { item => DocumentRef(item.index, item.`type`, item.id) })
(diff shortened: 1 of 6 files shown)


sksamuel:
Can you PR this against master (6,0) in its current form, which is great. For 5.4 though it will break backwards compatibility because you have changed more like this item to use a ref rather than index/type/id. I'd prefer not to do a breaking change in a bug release.

tyth:
What do you propose for 5.4 fix? Adding additional methods to work with `MoreLikeThisItemWithRouting` seems rather excessive.

sksamuel:
For 5.4, instead of changing

`case class MoreLikeThisItem(index: String, `type`: String, id: String)`

to

`case class MoreLikeThisItem(ref: DocumentRef, routing: Option[String] = None)`

you could just do

``case class MoreLikeThisItem(index: String, `type`: String, id: String, routing: Option[String] = None)`

And now its backwards compatible. Although the original change is better for 6.0 I think.

tyth:
Changed `MoreLikeThisItem` according to your last comment. This pull request is for 5.4.x, I will create one for 6.x later.

sksamuel:
👍

PR #946, opened 2017-06-11
Remove debug print from MoreLikeThisQueryBuilderFn

maxcom:
Debug print in this function add a lot of noise to our logs. This patch removes it.

Please also apply this to 5.4.x.


sksamuel:
:+1:

PR #947, opened 2017-06-12
handling FunctionScoreQuery, track_scores flag handled, sorting by _s…

mbarroux:
handling FunctionScoreQuery, track_scores flag handled, sorting by _score and GeoDistance

Commit initiated by "adrian83", I adapted for the 5.4.x version of elastic4s


sksamuel:
:+1:

mbarroux:
I would be glad if you could release this in 5.4.x version soon because we need it for our project, thank you :)

PR #952, opened 2017-06-19
Adding leading slash to IndexExists as in all other endpoints.

alex1712:
Hello,

As I was getting a BadRequest response trying to make an IndexExists call. I saw that the endpoint generated was something like this:
`https://search-h3se7w645ergom*****.eu-west-1.es.amazonaws.com:443index_name_bla`

So I found that in the IndexAdminImplicits the implementation for IndexExists was the only one without a leading '/'. I added it and now the end point is correctly generated like `https://search-h3se7w645ergom*****.eu-west-1.es.amazonaws.com:443/index_name_bla`.

I think that it is right now, but I wonder how nobody notice it before.

Thanks and I hope it helps :).


sksamuel:
Lovely thanks. Can you do same for master ?

Issue #954, opened 2017-06-19
Misleading naming of `DeleteByQueryDefinition.abortOnVersionConflict`


ptrlaszlo:
I was confused by the naming of `DeleteByQueryDefinition.abortOnVersionConflict`.

Based on the [code](https://github.com/sksamuel/elastic4s/blob/master/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/delete/DeleteImplicits.scala#L39-L41) deleting proceeds on conflict when `abortOnVersionConflict` is `true`, however in my opinion "Abort on version conflict = true" means aborting the operation when there is a conflict.

If you agree, I'll open a PR and rename `abortOnVersionConflict` to `proceedOnConflicts`.
Feel free to close this issue if you disagree.

sksamuel:
I think it was called that because that's what that Java API called it. I'm happy for it be renamed - please keep a forwarding method and deprecate it.

PR #955, opened 2017-06-19
Renaming abortOnVersionConflict to proceedOnConflicts

ptrlaszlo:
Renaming `abortOnVersionConflict` to `proceedOnConflicts` as agreed in https://github.com/sksamuel/elastic4s/issues/954


sksamuel:
awesome thanks

PR #956, opened 2017-06-21
Add support for NestedAggregationDefinition

bbroeksema:
Not much to add, I'm working on a project that is using NestedAggretations and as such need support for
it in elastic4s.


sksamuel:
:+1:

PR #959, opened 2017-06-23
Handle null value in SourceAsContentBuilder

fvlankvelt:
When a null value exists in the Map passed to the SourceAsContentBuilder, none of the types matches and a MatchError is thrown.  So handle this case explicitly.


sksamuel:
:+1:

PR #961, opened 2017-06-23
Add sub aggregation rather than replacing previously added ones

fvlankvelt:
The AggregationDefinition replaces all aggregations at once; particularly when using the singleton methods "subagg" and "subAggregation" this is unexpected.

This PR adds the provided sub aggregation to the list of aggregations already present in the definition.

The behavior of the multi-aggregate methods "subaggs" and "subAggregations" has been left unchanged, as expectations when those are invoked multiple times are not so clear.  Furthermore, the current behavior gives users the option of clearing the list of sub aggregations.

### DIFF
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/searches/aggs/AggregationDefinition.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/searches/aggs/AggregationDefinition.scala
@@ -11,11 +11,11 @@ trait AggregationDefinition extends AbstractAggregation {
 
   def subaggs: Seq[AbstractAggregation]
 
-  def subagg(agg: AbstractAggregation): T = subaggs(agg)
+  def subagg(agg: AbstractAggregation): T = subaggs(agg, subaggs: _*)
   def subaggs(first: AbstractAggregation, rest: AbstractAggregation*): T = subaggs(first +: rest)
   def subaggs(aggs: Iterable[AbstractAggregation]): T = subAggregations(aggs)
 
-  def subAggregation(agg: AbstractAggregation): T = subAggregations(agg)
+  def subAggregation(agg: AbstractAggregation): T = subAggregations(agg, subaggs: _*)
   def subAggregations(first: AbstractAggregation, rest: AbstractAggregation*): T = subAggregations(first +: rest)
   def subAggregations(aggs: Iterable[AbstractAggregation]): T
 }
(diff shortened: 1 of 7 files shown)


sksamuel:
Hmm, do you think builder methods should always append, or replace? Is it a matter of opinion? I agree it's not clear from the method name.

fvlankvelt:
Probably the clearest solutions would be to change the method names to addSubAggregation and addSubagg or to just provide the setters for the complete list.

Those solutions would change the signature though, which I would like to avoid.  The 'append' interpretation seemed natural to me.  The regular use case for builders is to add properties, with replacing being an exception.  So providing that option while making the regular case more powerful looked like the right thing to do.

sksamuel:
It's a question that was raised in another PR. I like the names to be free of add,set,append. But if you keep it like it is, then it's ambiguous. Someone else will raise a PR later saying they expect .subaggs to be a replacement as its an immutable data structure - if you do .copy in scala you don't get append, you get replace. So perhaps the addSubagg, withSubgagg, or setSubaggs is the best way.

bbroeksema:
I've been looking at this as well and was somehow under the impression that subaggs was append and not replace. Probably confused by the `first ++ rest` signatures. I have been looking around a bit and the `with*` naming seems to be more common in `scala(lala)-land`. So I'd suggest reverting the changes and add `withSubaggs` and `withSubAggregations` variants which add the desired functionality.

fvlankvelt:
Yeah, so keeping the .subaggs implementation as is does keep it in line with the .copy / replace interpretation.

The CustomNormalizerDefinition and CustomAnalyzerDefinition face a similar issue with the filters, resolving it with an 'addFilter' method name.  The 'with' name is used in conjunction with optional properties (type Option[]), so using it here could lead to further confusion.

bbroeksema:
@fvlankvelt I think you need to restore the old behavior of `subagg`/`subAggragation` and create a separate `add(Subagg|SubAggragation)` pair (and perhaps use `with` in stead of `add` as that seems more in line with common usage in scala).

sksamuel:
I think `add` is better than with, because of the simple reason when I first read the comment, I had to think for a second - was `with` adding, or was `with` replacing. I don't think it's _that_ clear in English, whereas add/append and set/replace are completely unambiguous.

I'd be in favour of deprecating the current and replacing them with set or update (or some other word), and adding `add` as this PR does. I'll merge this now.

PR #962, opened 2017-06-26
Add helpers for existance of aggragations

bbroeksema:
Calling any of the bucket or metric aggs functions on HasAggregations
with an invalid name, will result in a no-such-key exception. The
assumption seems to be that with an HasAggregation at hand, it is
always known which aggragations it will contain. This seems a bit
too strong an assumption, as such these helpers to figure out which
aggregations names are available.


sksamuel:
Good idea. However can we rename aggregations to something like names as it just returns the agg names and not the actual aggregations.

bbroeksema:
Good point, commit amended.

bbroeksema:
Ping?

sksamuel:
Great, thanks, merged.

PR #964, opened 2017-06-27
Dismax query builder would build an array in no field in the json object.

alex1712:
Adding field name to the array creation to avoid invalid json objects.


sksamuel:
Good spot thanks.

Issue #935, opened 2017-06-02
CreateIndexTemplate. Error mapping


fggarcia:
When execute

```
client.execute (
      createTemplate("hotels_metadata_template").pattern("hotels_metadata_*").mappings(
        mapping("metadata").fields(
          dateField("search_time"),
          dateField("metadata.checkin_date")
        )
      )
    ).await
```
output is:

> {"template":"hotels_metadata_*","mappings":{{"metadata":{"properties":{"search_time":{"type":"date"},"metadata.checkin_date":{"type":"date"}}}}}}

must be :

> {"template":"hotels_metadata_*","mappings":{"metadata":{"properties":{"search_time":{"type":"date"},"metadata.checkin_date":{"type":"date"}}}}}

I saw test but no implemented

elastic4s/elastic4s-tests/src/test/scala/com/sksamuel/elastic4s/indexes/CreateIndexTemplateDefinitionTest.scala

PR #968, opened 2017-06-28
Fix createTemplate json serialization

remcobeckers:
With release 6.0.0-M1 and 6.0.0-SNAPSHOT I ran into the issue where the `createTemplate` from the HttpClient threw an exception that an `ObjectNode` cannot be cast into an `ArrayNode`. I needed this in our own code so I made a fix that we are using now. I also implemented the test, although I could only get it to run for the HttpClient. There seems to be some kind of incompatibility for the TcpClient where that always fails.

Issue #935 might be related but it looks like it has a different cause that was already fixed in the 6.0.0 releases.


sksamuel:
Thank you.

PR #970, opened 2017-06-29
Fix rawField deprecation warnings

solicode:
Fixes `rawField` deprecation warnings and avoids content type auto-detection.


sksamuel:
Great thanks

PR #975, opened 2017-06-30
Added parent to SearchHit

### LINKED ISSUE #974: _parent field is missing from com.sksamuel.elastic4s.http.search.SearchHit

douglaz:
Fixes #974


sksamuel:
:+1:

PR #976, opened 2017-07-03
Analysers in template

remcobeckers:
I noticed that analyzers defined in a template were not included in the JSON that was submitted by the HttpClient. I created a fix for that and, to be able to test it easily, also added a Show typeclass for the CreateIndexTemplateDefinition type.


remcobeckers:
Looks like rebasing on master broke my fix again. I'll have a look later today. Also my fix for `matchJson` apparently started braking some other tests. I'll check this out too.

sksamuel:
The tests are flaky on master. As long as it works locally I'll merge it.

remcobeckers:
There were indeed a few regressions that I fixed now.

sksamuel:
:+1:

PR #978, opened 2017-07-04
Fixing `HitReader Typeclass` documentation.

roberttaylor426:
Correct me if I'm wrong, but this documentation is out-of-date / wrong.


sksamuel:
Thanks for fixing.

PR #996, opened 2017-07-24
support for CumulativeSum and DateRange in HTTP aggs

dankins:
thank you for the work you've done on this project! 

This PR adds support for DateRange and CumulativeSum aggregations in the http API.


sksamuel:
Great thanks. If you want them in 6.0.0 too, then you could PR against master too, the format has changed slightly in the responses.

PR #1004, opened 2017-07-28
Fix MatchError: null

tksfz:


sksamuel:
I can't merge this as there's a conflict (probably caused by the fix I did for it earlier).

tksfz:
Excellent

PR #1008, opened 2017-07-30
Add  field for Highlight Bulders

generall:
It is possible now to use 'options' filed for highlighting.
It is necessary, e.g. for https://github.com/wikimedia/search-highlighter plugin


sksamuel:
:+1:

PR #1014, opened 2017-08-02
Always request version on search

douglaz:
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-version.html
I see no reason to leave it off, specially given that `version` on `SearchHit` is non-optional.


sksamuel:
There must be a reason it's optional though?

douglaz:
@sksamuel without the version, it'll save some bytes on each hit response. The version can be a number with ~18 digits (`The value provided must be a numeric, long value greater or equal to 0, and less than around 9.2e+18`)

Besides the response size, it could be slower, but  query time seems to be identical for practical purposes.

sksamuel:
:+1:

PR #1015, opened 2017-08-03
Scala 2.12 for 2.4

### LINKED ISSUE #957: Publish 2.12 version of lib for elasticsearch 2.x
Hi,

We're using this client against an ES cluster that is running ES versions 2.x. I noticed that you only published 2.12 version for 5.x+. Since ES version 2.x is not EOL and still very much used in production, can you publish 2.12 for that as well? Thanks!

avdv:
Fixes #957.

Would you want to have the commits squashed together?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,9 @@ before_script:
 - ulimit -u 30000
 scala:
 - 2.11.8
-- 2.10.6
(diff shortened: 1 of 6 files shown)


sksamuel:
:+1:

PR #1017, opened 2017-08-05
Fix inner hits propagation from HasChildren builder to JSON query

kurochenko:
This PR fixes propagation of `inner_hits` attribute of `has_child` query. Example:

```
val query = search("myidx").query(
    hasChildQuery("currencies")
        .query(matchQuery("referential", true))
        .scoreMode(ScoreMode.None)
        .innerHit(InnerHitDefinition("currencies"))
)
client.show(query)
```
would result into 
```
{
  "query": {
    "has_child": {
      "type": "currencies",
      "score_mode": "none",
      "query": {
        "match": { "referential": { "query": true } }
      }
    }
  }
}
```
Where `inner_hits` attribute is missing.

Also `name` attribute of `inner_hits` was not propagated. Fixed along the way in this PR as the two problems are tightly coupled.


sksamuel:
:+1:

PR #1029, opened 2017-08-17
Make Jackson deserialization work proper for some generics

douglaz:
Warning: this change breaks this test (but perhaps not the feature being tested):
```scala
"support custom mapper" in {

      implicit val mapper: ObjectMapper = mock[ObjectMapper]
      Mockito.when(mapper.readTree(org.mockito.Matchers.anyString)).thenReturn(mock[ObjectNode])

      val resp = client.execute {
        search("jacksontest" / "characters").query("breaking")
      }.await
      // our custom mapper will just return null so that should be returned
      resp.to[Character].toList shouldBe List(null)
    }
```

This fixes serialization of objects shaped like `SomeType[Generic1, Generic2]` because `ScalaObjectMapper.constructType[T]` handle all corner cases while `manifest.runtimeClass.asInstanceOf[Class[T]]` is too simplistic.

### DIFF
--- a/elastic4s-jackson/src/main/scala/com/sksamuel/elastic4s/jackson/ElasticJackson.scala
+++ b/elastic4s-jackson/src/main/scala/com/sksamuel/elastic4s/jackson/ElasticJackson.scala
@@ -1,8 +1,8 @@
 package com.sksamuel.elastic4s.jackson
 
-import com.fasterxml.jackson.annotation.JsonFormat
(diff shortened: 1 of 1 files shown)


sksamuel:
:+1:

sksamuel:
Don't worry about the tests, I'm trying to make them all work after upgrading to 6.0-beta1 when have the stuff has changed again :\

PR #1034, opened 2017-08-21
implement exponential decay, weight and linear decay in http client

kharnt0x:
add test json for fieldvaluefactor
add multi_value_mode to supported scorers based on elastic doco. Using uppercase for the enum values to share test data with TCP
add weight to supported scorers as per elastic doco
add enum conversion for MultiValueMode (except "median" which doesn't exist in doco and causes an error in Elastic 5.x +)


sksamuel:
Commit looks fantastic so far. Let me know when you're done pushing and I'll merge when green.

kharnt0x:
@sksamuel I just fixed the tests again, should be all green and ready to merge

sksamuel:
Brilliant thanks.

Issue #1044, opened 2017-08-28
inner_hits for has_parent isn't propagated in http client 


ingarabr:
Same issue as described in https://github.com/sksamuel/elastic4s/pull/1017 but for `has_parent`.

PR #1046, opened 2017-08-28
Inner hits for has parent 5.4.x

ingarabr:
Backport for 5.4.

https://github.com/sksamuel/elastic4s/issues/1044

Was thinking about backporting https://github.com/sksamuel/elastic4s/pull/1017 since they're related.


sksamuel:
:+1:

Issue #1047, opened 2017-08-28
Filter not added correctly to function score scorers when using http client


kharnt0x:
The following function score query:
```
functionScoreQuery()
      .scoreFuncs(
filterFunction(weightScore(100)).filter(matchPhraseQuery("foo", "bar"))
)
```
should result in something like:
```
{
    "function_score": {
        "functions": [
            {
                "weight": 100,
                "filter": {
                    "match_phrase": {
                        "foo": {
                            "query": "bar"
                        }
                    }
                }
            }
        ]
    }
}
```
instead we just get:
```
{
    "function_score": {
        "functions": [
            {
                "weight": 100
            }
        ]
    }
}
```
Not only does it not work, the API is very confusing. Elasticsearch allows "filter" to be added to all the Score functions just like a weight can be. Right now however there is a confusing type `FilterFunctionDefinition` that must be used in order to add a filter.
I'm proposing to add an optional filter property to each ScoreFunctionDefinition and to remove the type FilterFunctionDefinition altogether. 

I'll create a PR and let me know what you think please @sksamuel

kharnt0x:
@sksamuel I've created https://github.com/sksamuel/elastic4s/pull/1048 which is a simple fix for the above issue. I'll look at changing the API to be more consistent with a separate PR

sksamuel:
I've merged the simple fix for now. The reason for the confusing API is because it's a copy of the way the Java API works. I agree it's not very good. Feel free to propose something completely different.

kharnt0x:
@sksamuel I've created https://github.com/sksamuel/elastic4s/pull/1049

Let me know what you think

sksamuel:
Closing this as the new API is very nice.

PR #1049, opened 2017-08-29
Simplify function score API

kharnt0x:
simplifying API as per https://github.com/sksamuel/elastic4s/issues/1047
Also added tests for function_score query as a whole when using TCP client

### DIFF
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/searches/queries/funcscorer/FunctionScoreQueryDefinition.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/searches/queries/funcscorer/FunctionScoreQueryDefinition.scala
@@ -3,13 +3,12 @@ package com.sksamuel.elastic4s.searches.queries.funcscorer
 import com.sksamuel.elastic4s.searches.queries.QueryDefinition
 
 case class FunctionScoreQueryDefinition(query: Option[QueryDefinition] = None,
-                                        scorers: Seq[ScoreFunctionDefinition] = Nil,
+                                        functions: Seq[ScoreFunctionDefinition] = Nil,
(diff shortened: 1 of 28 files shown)


elastic4s-core/src/main/scala/com/sksamuel/elastic4s/searches/queries/funcscorer/FunctionScoreQueryDefinition.scala
```diff
@@ -29,10 +29,10 @@ case class FunctionScoreQueryDefinition(query: Option[QueryDefinition] = None,
               rest: ScoreFunctionDefinition*): FunctionScoreQueryDefinition = scorers(first +: rest)
 
   def scorers(scorers: Iterable[ScoreFunctionDefinition]): FunctionScoreQueryDefinition =
-    scoreFuncs(scorers.map(FilterFunctionDefinition(_)))
+    scoreFuncs(scorers)
 
-  def scoreFuncs(first: FilterFunctionDefinition,
```
kharnt0x: @sksamuel  we could remove `scoreFuncs` altogether now and just leave `scorers`

sksamuel:
Overall very good. One comment - should we rename scorers to functions, since it's called functions in the json and that's what people will see when they read the elasticsearch docs. Then we can remove scoreFuncs as you say as well. It was called scorers in the code as a left over from the Java TCP API. 

Also, `scriptScore` is a proper field in the `FunctionScoreQueryDefinition` but it could be removed as it's a score function anyway.

kharnt0x:
@sksamuel good suggestions, this should help with discoverability for sure. I've made the updates

sksamuel:
Sorry to be a pain, can we keep the original methods just with deprecation. I'll merge this if you want to do that as a separate pr.

Issue #1003, opened 2017-07-28
GeoBounding box crashing because of com.fasterxml.jackson.databind.node.ObjectNode cannot be cast to com.fasterxml.jackson.databind.node.ArrayNode on GeoBoundingBoxQueryBodyFn


SodaDev:
Start object is called too many times during query object creation which causes throwing an exception. Fix for that issue is very simple one. Since I have no access to the repo, I've attached simple patch for that fix. You can check that behaviour in the REPL by calling `XContentFactory.jsonBuilder().startObject().startObject()` which causes 
```
com.fasterxml.jackson.databind.node.ObjectNode cannot be cast to com.fasterxml.jackson.databind.node.ArrayNode
java.lang.ClassCastException: com.fasterxml.jackson.databind.node.ObjectNode cannot be cast to com.fasterxml.jackson.databind.node.ArrayNode
	at com.sksamuel.elastic4s.json.XContentBuilder.array(XContentBuilder.scala:22)
	at com.sksamuel.elastic4s.json.XContentBuilder.startObject(XContentBuilder.scala:209)
	at com.sksamuel.elastic4s.search.SearchDslTest$$anonfun$79.apply(SearchDslTest.scala:462)
	at com.sksamuel.elastic4s.search.SearchDslTest$$anonfun$79.apply(SearchDslTest.scala:462)
	at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
	at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
	at org.scalatest.Transformer.apply(Transformer.scala:22)
	at org.scalatest.Transformer.apply(Transformer.scala:20)
	at org.scalatest.FlatSpecLike$$anon$1.apply(FlatSpecLike.scala:1682)
	at org.scalatest.TestSuite$class.withFixture(TestSuite.scala:196)
	at org.scalatest.FlatSpec.withFixture(FlatSpec.scala:1685)
	at org.scalatest.FlatSpecLike$class.invokeWithFixture$1(FlatSpecLike.scala:1679)
	at org.scalatest.FlatSpecLike$$anonfun$runTest$1.apply(FlatSpecLike.scala:1692)
	at org.scalatest.FlatSpecLike$$anonfun$runTest$1.apply(FlatSpecLike.scala:1692)
	at org.scalatest.SuperEngine.runTestImpl(Engine.scala:289)
[...]

PR #1052, opened 2017-08-30
Fix for Geo Box query builder

SodaDev:
Solving the issue related to the Issue: https://github.com/sksamuel/elastic4s/issues/1003


sksamuel:
👍

Issue #914, opened 2017-05-15
CompletionSuggestionDefinition is not matched inside http client while parsing the response


oguzalbayrakDH:
Hello,
CompletionSuggestionDefinition is not matched inside http client while parsing the response, so when the user tries to parse a completionSuggestion response a MatchError is raised from the http client
This is the library code inside SearchBodyBuilderFn.scala that causes the problem
```
    if (request.suggs.nonEmpty) {
      builder.startObject("suggest")
      request.suggs.foreach {
        case term: TermSuggestionDefinition =>
          builder.startObject(term.name)
          term.text.foreach(builder.field("text", _))
          builder.startObject("term")
          builder.field("field", term.fieldname)
          term.analyzer.foreach(builder.field("analyzer", _))
          term.lowercaseTerms.foreach(builder.field("lowercase_terms", _))
          term.maxEdits.foreach(builder.field("max_edits", _))
          term.minWordLength.foreach(builder.field("min_word_length", _))
          term.maxInspections.foreach(builder.field("max_inspections", _))
          term.minDocFreq.foreach(builder.field("min_doc_freq", _))
          term.maxTermFreq.foreach(builder.field("max_term_freq", _))
          term.prefixLength.foreach(builder.field("prefix_length", _))
          term.size.foreach(builder.field("size", _))
          term.shardSize.foreach(builder.field("shard_size", _))
          term.sort.map(_.name().toLowerCase).foreach(builder.field("sort", _))
          term.stringDistance.map(_.name.toLowerCase).foreach(builder.field("string_distance", _))
          term.suggestMode.map(_.name().toLowerCase).foreach(builder.field("suggest_mode", _))
          builder.endObject()
          builder.endObject()
      }
      builder.endObject()
    }
```

this is what i got using completionSuggestion in a query

```
play.api.http.HttpErrorHandlerExceptions$$anon$1: Execution exception[[MatchError: CompletionSuggestionDefinition(mysugg,title,None,None,None,None,None,None,None,None,None,None,None,None,Some(/)) (of class com.sksamuel.elastic4s.searches.suggestion.CompletionSuggestionDefinition)]]
	at play.api.http.HttpErrorHandlerExceptions$.throwableToUsefulException(HttpErrorHandler.scala:293)
	at play.api.http.DefaultHttpErrorHandler.onServerError(HttpErrorHandler.scala:220)
	at play.api.GlobalSettings$class.onError(GlobalSettings.scala:160)
[...]

PR #1056, opened 2017-08-31
Added completion and phrase suggestion support

MatteCarra:
I've added support for completion and phrase suggestion (issue #914 )
This patch is not completed yet because I've not been able the documentation for some fields (see TODOs in CompletionSuggestionBuilderFn and  PhraseSuggestionBuilderFn).
I will be working some more on this as soon as possible.
I've done some tests only for completion suggestion


sksamuel:
Looks good.

PR #1057, opened 2017-09-01
Completion and Phrase suggestion

MatteCarra:
I've completed all todos and I've tested some more completion suggestion.
It would be nice if someone tests phrase suggestion and gives some feedback.


sksamuel:
:+1:

PR #1062, opened 2017-09-04
Completion context mapping (Port from release/5.4.x)

MatteCarra:
I've also added some convertions that I forgot to do


sksamuel:
👍

PR #1064, opened 2017-09-04
timezone in date range fix

### LINKED ISSUE #1063: DateRange aggregation ignores timeZone
DateRangeBuilder actually ignores timezone that was set in DateRangeAggregation.

maxcom:
fixes #1063


sksamuel:
👍

PR #1068, opened 2017-09-06
Update by query

### LINKED ISSUE #1006: No HttpExecutable for UpdateByQueryDefinition
I'm using elastic4s 5.4.5 with Http client.

When I try to execute UpdateByQueryDefinition,
it fails with:

> could not find implicit value for parameter exec: com.sksamuel.elastic4s.http.HttpExecutable[com.sksamuel.elastic4s.update.UpdateByQueryDefinition,U]

After looking some codes in elastic4s-http module,
I could not find implicits for UpdateByQueryDefinition.
It seems to have implicits only for UpdateDefinition.

Is there something I missed?

kbedel:
Resolves #1006 for 6.x. If this looks good, I will also submit a PR for the 5.5.x branch

### DIFF
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/update/UpdateApi.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/update/UpdateApi.scala
@@ -1,7 +1,7 @@
 package com.sksamuel.elastic4s.update
 
 import com.sksamuel.elastic4s.searches.queries.QueryDefinition
-import com.sksamuel.elastic4s.{IndexAndTypes, Indexes}
+import com.sksamuel.elastic4s.{IndexAndTypes, Indexes, IndexesAndTypes}
 
 trait UpdateApi {
 
@@ -10,9 +10,11 @@ trait UpdateApi {
     def in(indexType: IndexAndTypes): UpdateDefinition = UpdateDefinition(indexType, id.toString)
   }
 
-  def updateIn(indexes: Indexes): UpdateExpectsQuery = new UpdateExpectsQuery(indexes)
-  class UpdateExpectsQuery(indexes: Indexes) {
-    def query(query: QueryDefinition) = UpdateByQueryDefinition(indexes, query)
+  def updateIn(indexes: Indexes): UpdateExpectsQuery = new UpdateExpectsQuery(indexes.toIndexesAndTypes)
+  def updateIn(indexesAndTypes: IndexesAndTypes): UpdateExpectsQuery = new UpdateExpectsQuery(indexesAndTypes)
+
+  class UpdateExpectsQuery(indexesAndTypes: IndexesAndTypes) {
+    def query(query: QueryDefinition) = UpdateByQueryDefinition(indexesAndTypes, query)
   }
 }
 
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/update/UpdateApi.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/update/UpdateApi.scala
@@ -11,6 +11,8 @@ trait UpdateApi {
   }
 
   def updateIn(indexes: Indexes): UpdateExpectsQuery = new UpdateExpectsQuery(indexes.toIndexesAndTypes)
+
+  @deprecated("Use updateIn(index) which only supports updating by index, not by index and type. This is because in version 6 of Elasticsearch types are deprecated https://www.elastic.co/guide/en/elasticsearch/reference/master/removal-of-types.html.", "6.0.0")
   def updateIn(indexesAndTypes: IndexesAndTypes): UpdateExpectsQuery = new UpdateExpectsQuery(indexesAndTypes)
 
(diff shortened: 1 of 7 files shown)


elastic4s-core/src/main/scala/com/sksamuel/elastic4s/update/UpdateByQueryDefinition.scala
```diff
@@ -1,13 +1,13 @@
...
-import com.sksamuel.elastic4s.Indexes
+import com.sksamuel.elastic4s.IndexesAndTypes
 import com.sksamuel.elastic4s.script.ScriptDefinition
 import com.sksamuel.elastic4s.searches.queries.QueryDefinition
 import com.sksamuel.exts.OptionImplicits._
 
 import scala.concurrent.duration.FiniteDuration
 
-case class UpdateByQueryDefinition(sourceIndexes: Indexes,
+case class UpdateByQueryDefinition(indexesAndTypes: IndexesAndTypes,
```
kbedel: Are breaking changes in these inner query definitions OK? What about at minor released as well if I merge it to 5.5.x branch?

elastic4s-core/src/main/scala/com/sksamuel/elastic4s/update/UpdateApi.scala
```diff
@@ -10,9 +10,11 @@ trait UpdateApi {
     def in(indexType: IndexAndTypes): UpdateDefinition = UpdateDefinition(indexType, id.toString)
   }
 
-  def updateIn(indexes: Indexes): UpdateExpectsQuery = new UpdateExpectsQuery(indexes)
-  class UpdateExpectsQuery(indexes: Indexes) {
-    def query(query: QueryDefinition) = UpdateByQueryDefinition(indexes, query)
+  def updateIn(indexes: Indexes): UpdateExpectsQuery = new UpdateExpectsQuery(indexes.toIndexesAndTypes)
+  def updateIn(indexesAndTypes: IndexesAndTypes): UpdateExpectsQuery = new UpdateExpectsQuery(indexesAndTypes)
```
kbedel: If you don't think it makes sense to add this in master for 6.x, I'm fine with removing this, but I would like to include it in the 5.5.x branch

elastic4s-core/src/main/scala/com/sksamuel/elastic4s/update/UpdateByQueryDefinition.scala
sksamuel: I'm ok with this changing, but lets add an apply method in a companion object that accepts Indexes and converts it, so it doesn't break anything.

elastic4s-core/src/main/scala/com/sksamuel/elastic4s/update/UpdateApi.scala
sksamuel: I think we can add both, but in 6.0 lets immediately deprecate it. This means people who need to update by type can do so, but they'll get the warning that it's going away in 7.

sksamuel:
Great. Once this goes green I'll merge.

elastic4s-core/src/main/scala/com/sksamuel/elastic4s/update/UpdateByQueryDefinition.scala
```diff
@@ -30,6 +31,8 @@ case class UpdateByQueryDefinition(sourceIndexes: Indexes,
 
   def refresh(refresh: Boolean): UpdateByQueryDefinition = copy(refresh = refresh.some)
```
kbedel: I noticed that DeleteByQueryDefinition uses a RefreshPolicy Enum for this, and now sets the query parameter based on it. Would you like me to make the equivalent change here?

sksamuel: Please.

elastic4s-tcp/src/main/scala/com/sksamuel/elastic4s/update/UpdateExecutables.scala
```diff
@@ -26,7 +26,8 @@ trait UpdateExecutables {
       builder.filter(QueryBuilderFn(t.query))
       t.requestsPerSecond.foreach(builder.setRequestsPerSecond)
       t.maxRetries.foreach(builder.setMaxRetries)
-      t.refresh.foreach(builder.refresh)
+      if (t.refresh.exists(_ == RefreshPolicy.Immediate))
+        builder.refresh(true)
```
sksamuel: This is problematic because if someone choose wait, then it won't really wait until it's finished.
Maybe we should throw an exception.

kbedel: OK, I just duplicated the logic that was in DeleteExecutables. Leaving this as-is will keep those in sync. Would fixing both of those be something that would be better handled in it's own issue?

sksamuel: Yes, good idea.

PR #1069, opened 2017-09-06
Update by query 5.x

kbedel:
5.5.x version of fix for #1006


elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/delete/DeleteImplicits.scala
```diff
@@ -39,6 +39,7 @@ trait DeleteImplicits {
       if (request.abortOnVersionConflict.contains(true)) {
         params.put("conflicts", "proceed")
       }
+      request.refresh.map(RefreshPolicyHttpValue.apply).foreach(params.put("refresh", _))
```
kbedel: This isn't strictly related to the issue, but noticed it was missing in 5.5.x when adding it to UpdateImplicits so added it here as well.

sksamuel: Good spot.

PR #1073, opened 2017-09-07
add travis cache

kimxogus:
I found that travis cache is not configured. This will reduce travis build time.

[sbt doc](http://www.scala-sbt.org/0.13/docs/Travis-CI-with-sbt.html)


sksamuel:
Awesome.

PR #1078, opened 2017-09-09
Url parameter

kimxogus:
- Fix `ReindexContextBuilder`
- Added `AbstractURLParameterDefinition` for url parameters in `HttpClient` request.
- Added `asyncTimeout` to set timeout for async request.


sksamuel:
Great PR, what about making AbstractURLParameterDefinition a case class, and just having it as a field in any of the requests that need it?

kimxogus:
Well, as I know, extending case class is not possible.
I added BaseURLParameterDefinition case class instead.

sksamuel:
I mean like have the case class

`case class UrlParameters`

and then in each request

`case class SearchRequest(index:String, ....., urlParams: UrlParameters, ....)`

kimxogus:
Ah, that makes sense.
Looks good. I'll change it to use case class.

sksamuel:
Then you don't need to define the fields in each request.

sksamuel:
We'll still need the .timeout() .refresh() methods etc, which will just do .copy(url = url.copy(...))

kimxogus:
Changes applied.

sksamuel:
I think it looks great but one more change :)
I don't think the url parameters itself needs to be Optional, because all the values inside it are optional.

kimxogus:
Updated as requested :)

sksamuel:
Awesome.

Issue #1075, opened 2017-09-07
Empty GetIndexTemplateResponse


kimxogus:
[GetIndexTemplateResponse](https://github.com/sksamuel/elastic4s/blob/2f210e55c4fa407bf4daef6b108a13dd43272836/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/index/IndexTemplateImplicits.scala#L15) is empty case class, so I can't retrieve index template.
Also, response for create, delete index template are empty. They need to be implemented.

sksamuel:
I probably started it and didn't get around to it yet. It'll be in 6.0.0 final of course.

sksamuel:
This is now in master and will be part of 6.0.0-rc2.

PR #1079, opened 2017-09-09
add IndexTemplateResponse properties

kimxogus:
Just temporal fix for #1075 
Is this ok?


kimxogus:
Updated to use case class

PR #1083, opened 2017-09-25
ElastichsearchClientUri: Accept trailing slash in connection string

shanielh:


sksamuel:
Great thanks.

shanielh:
Thanks for merging so fast :) Do you know when is the next release? :smile:

Issue #1085, opened 2017-09-25
Updating the index of a document with indexInto fails


yanns:
The following code:
```
client.execute { indexInto(indexName) doc document id document.id }
```
fails if the document already exists.
Elatisc Search returns a 200 status code in that case, that is not handle in [IndexImplicits](https://github.com/sksamuel/elastic4s/blob/5210868d48498a1161b066a06d4900e671c4120e/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/index/IndexImplicits.scala#L21):

```
    override def responseHandler: ResponseHandler[Either[IndexFailure, IndexResponse]] = new ResponseHandler[Either[IndexFailure, IndexResponse]] {
      override def doit(response: HttpResponse): Either[IndexFailure, IndexResponse] = response.statusCode match {
        case 201 => Right(ResponseHandler.fromEntity[IndexResponse](response.entity.getOrError("Index responses must have a body")))
        case 400 | 500 => Left(IndexFailure(response.entity.get.content))
        case _ => sys.error(response.toString)
      }
    }
```

PR #1086, opened 2017-09-25
Fix 1085: handle 200 response status code when indexing

yanns:
When indexing an existing document (with the same id), elastic search
answers with a 200 response status.

fix https://github.com/sksamuel/elastic4s/issues/1085


sksamuel:
Can you add a test so I don't get caught out again. Can be added to the existing index test suite.

yanns:
Yes, I tried to do that, but have not really found an existing test that would fit this change.

Do you have any pointer of existing tests I could use as a "template"?

sksamuel:
com.sksamuel.elastic4s.indexes.IndexTest

All you need to do is add another that indexes the same document twice I think ? First time it will be 201 second time 200.

yanns:
Thx for the pointer!

I added a test: https://github.com/sksamuel/elastic4s/pull/1086/commits/0362f6626204ebeab0e6fd6b5d52442729c40f34

Do you prefer several commits? Or a rebase?

sksamuel:
Perfect I'll merge once green.

PR #1089, opened 2017-09-26
Add compound word token filter to DSL

jadlr:
Thanks for this great library!

I found that this was missing from the DSL so added it. The documentation can be found [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-compound-word-tokenfilter.html)


sksamuel:
:+1:

PR #1090, opened 2017-09-26
Do not explicitly set defaults in shingle token filter

jadlr:
I think it is not a good idea to set defaults in the DSL. If these defaults will change in future ES versions they will be overwritten which is most likely not what the user wants.

What do you think?


sksamuel:
Yes, that's how 95% of the rest of the DSL is, and anything with a default is a hold over from a few years ago. Feel free to change any!

PR #1092, opened 2017-09-26
Add InnerHit.sort property for inner_hits sort calculations

sitio:
An example of where this is useful is in retrieving the geo distance of a nested object to determine the actual distance from a queried/filtered point.

This in an inner_hits filter:

`
    "sort": [
        {
            "_geo_distance": {
                "mode": "min",
                "offices.geo": [
                    [
                        -118.2735595703125,
                        34.063804626464844
                    ]
                ],
                "order": "asc"
            }
        }
    ]`

Will produce a "sort" property of the inner_hits json that contains the distance of that specific office. 

If nothing else, it is something returned by Elastic Search... it seems like a good thing to have in the elastic4s case class.

Thank you for your consideration.


sksamuel:
:+1:

PR #1095, opened 2017-09-27
Remove more hard-coded defaults from TokenFilter

jadlr:
There are still some left, I'll try to remove them once I have time.


sksamuel:
:+1:

Issue #1098, opened 2017-10-03
QueryStringQueryDefinition do not provide a way to set useAllFields property of java client QueryStringQueryBuilder (all_fields property in Elasticsearch Query API)) )


devsprint:
QueryStringQueryDefinition do not provide a way to set useAllFields property of java's client QueryStringQueryBuilder (all_fields property in Elasticsearch Query API)

One usage of this feature is Highlighting feature where you want to highlight values based on a search query but the list of fields could not be determined at the time of the request.

sksamuel:
Which version?

devsprint:
5.4.x, I did not found any trace of support for this field in source code. I also checked the master source code.

sksamuel:
Yeah it's been removed in 6.0, `The all_fields parameter for the query_string has been removed. Set default_field to *` instead. If provided, default_field will be automatically set to *`
So there's no point adding it I don't think.

https://www.elastic.co/guide/en/elasticsearch/reference/6.0/breaking_60_search_changes.html

devsprint:
Well, I have seen that as being deprecated in Java library too, but setting only the default_field to * is not really working, because it will try to match on fields of other type like Numeric fields. Lenient property may remove the errors but I'm not sure if this  approach is not having other side effects...

sksamuel:
Hmm if the approach is invalid for numbers, might be worth creating an issue with elastic before 6.0 is released.

sksamuel:
Or if you to add it for 5.4 branch, if you do a PR I'll release it.

devsprint:
Ok,  I will try to not expose that field out of QueryStringQueryDefinition because it will break backward compatibility,case class will have 23 parameters in constructor, as I have seen that code is compiled for older version of scala. I will do it in a way to not break it.

sksamuel:
You could put it on one of the "child" case classes of the QueryStringQueryDefinition case class.

PR #1099, opened 2017-10-03
all_fields support in query string without breaking API, #1098

devsprint:
Not the ideal implementation but it is least intrusive than other solutions. It will set all_fields to true only in case there the query is not including any fields and there is no default_field defined.


sksamuel:
:+1:

devsprint:
@sksamuel is there anything that I should do to make the Travis CI to run the build? It looks like having issues with git config.

sksamuel:
Looks fine https://travis-ci.org/sksamuel/elastic4s/builds/282708686

devsprint:
@sksamuel Thank you. I will look for next release on 5.4.x.

Issue #1100, opened 2017-10-03
SpanNearQueryBodyFn generates SpanOrQueryBodyFn instead


unkarjedy:
This bug affects all versions.
see SpanNearQueryBodyFn.scala:
`builder.startObject("span_or")`
I believe this is caused by copypasting.
The fix is tiny: just replace `span_or` with `span_near` in all versions.

I cathed it in  5.4 version which actually contains another bug in above classes:
`builder.rawValue` inside an array does not add any comma between array entries.

PR #1101, opened 2017-10-03
#1100 Fixed SpanNearQueryBIdyFn.scala to generate span_near ES query …

unkarjedy:
This bug affects all versions.
see SpanNearQueryBodyFn.scala:
builder.startObject("span_or")
I believe this is caused by copypasting.
The fix is tiny: just replace span_or with span_near in all versions.

I cathed it in 5.4 version which actually contains another bug in above classes:
builder.rawValue inside an array does not add any comma between array entries.


sksamuel:
:+1:

PR #1102, opened 2017-10-04
Fixed crash during building sorting by GeoDistance

zlumyo:
Structure of builder should be similar to ScoreSort and starts with empty object


zlumyo:
This is an old issue, going from many older releases.

sksamuel:
:+1:

PR #1103, opened 2017-10-04
Fix limit token count token filter

jadlr:
The filter was using the type 'stop' instead of 'limit'. See the [docs](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/analysis-limit-token-count-tokenfilter.html)


sksamuel:
:+1:

PR #1104, opened 2017-10-04
Remove redundant token filter

jadlr:
The functionality that the removed `StopTokenFilterPath` was implementing can be achieved using the `StopTokenFilter` as can be seen in `CreateIndexApiTest`.


sksamuel:
:+1:

PR #1105, opened 2017-10-04
Remove hard-coded defaults from token filters

jadlr:
This should remove the rest of the hard-coded defaults.


sksamuel:
Really nice work @jadlr

PR #1107, opened 2017-10-05
1100 span near produces span or query (copy from 5.4 branch)

unkarjedy:
Copy of pull request-1101 (was merged into 5.4)
https://github.com/sksamuel/elastic4s/pull/1101

### DIFF
--- a/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/search/queries/span/SpanNearQueryBodyFn.scala
+++ b/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/search/queries/span/SpanNearQueryBodyFn.scala
@@ -1,6 +1,7 @@
 package com.sksamuel.elastic4s.http.search.queries.span
 
 import com.sksamuel.elastic4s.http.search.queries.QueryBuilderFn
+import com.sksamuel.elastic4s.http.search.queries.span.XContentBuilderExtensions._
 import com.sksamuel.elastic4s.searches.queries.span.SpanNearQueryDefinition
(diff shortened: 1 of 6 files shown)


sksamuel:
This is bringing in the all_fields change as well as the span change.

unkarjedy:
@sksamuel  Sorry... Fixed. 
Could you please suggest me how to apply this fix to all release branches automatically?
Or at least semi-automatically?

sksamuel:
The closest you will get is `git cherry-pick`, are you familiar with that?

unkarjedy:
@sksamuel 
Yes, that is the way I do it right now... 
Actually, I am quite good at Git, but I have little experience with its Web wrappers like GitHub. 
I thought that maybe there are some tools to auto-create several pull-requests.
I think my time has come and I should create such tool =D

sksamuel:
Yes could be a github plugin.

sksamuel:
Thanks for the PR. Can you explain this rawArray method to me, I didn't understand why you added it.

unkarjedy:
@sksamuel  
For some reason, `XContentBuilder` did not write a comma after each clause object bytes was added with `writeRaw` method. I've spent enough time reverse-engineering XContentBuilder and did not find the `normal` way to fix it. As a workaround, I collect all the raw values in one buffer and insert comma byte between them. After that, I write one big raw value as raw array value.

As I understood you have replaced elsticsearch `XContentBuilder` with your own implementation in version 6.x. And there is no such bug there.

sksamuel:
Oh it's in a bug in 'their' version. Ok that's why I was confused, I thought it was a bug in mine.
Good work.

PR #1109, opened 2017-10-05
Use "source" instead of the deprecated "inline" key for scripts

jonaslan:
Hi!
I'm getting a bunch of warnings when using ES 5.6. Apparently, the "inline" key will no longer be supported for ES 6.0. 
https://github.com/elastic/elasticsearch/issues/26823

Note that it breaks compatibility with ES 5.5.

What do you think? Any feedback is appreciated, and thanks for maintaining this!

### DIFF
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/script/ScriptDefinition.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/script/ScriptDefinition.scala
@@ -4,7 +4,7 @@ import scala.language.implicitConversions
 
 case class ScriptDefinition(script: String,
                             lang: Option[String] = None,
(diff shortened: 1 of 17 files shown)


sksamuel:
I would deprecate ScriptType.Inline and introduce ScriptType.Source, then make inline point to source.
Then the API will match up with what ES expects.
Thanks for the PR :+1:

sksamuel:
:+1:

jonaslan:
I don't know if this is what you meant, but now ScriptType.Source is default and ScriptType.Inline is deprecated. Both are still supported though. (should have done it this way from the beginning, thanks for pointing me in the right direction)

sksamuel:
Yep that's what I meant.

PR #1117, opened 2017-10-06
(copy from 5.4 to 5.6) Fixed SpanNearQueryBIdyFn.scala to generate span_near ES query instead of span_or query

unkarjedy:
(copy from 5.4 to 5.6) Fixed SpanNearQueryBIdyFn.scala to generate span_near ES query instead of span_or query


sksamuel:
:+1:

PR #1118, opened 2017-10-06
(copy from 5.4 to master) Fixed SpanNearQueryBIdyFn.scala to generate span_near ES query instead of span_or query 

unkarjedy:
did not include XContentBuilderExtentions due to this bug is not present in master branch


sksamuel:
:+1:

PR #1120, opened 2017-10-17
FiltersAggregations and KeyedFiltersAggregations for http

jonaslan:
I noticed that FiltersAggregations were missing from the http client. These fill my needs for the moment, but I'm not 100% confident with aggregations so tell me if there's something I missed.

I wasn't sure whether to use named or unnamed filters for the keyed filters. Which would you prefer?


sksamuel:
Looks great.

PR #1121, opened 2017-10-17
Fix SpanFirstQueryBodyFn

catalin-ursachi:
This was throwing an exception, because the writer was closing one object too many. But also, the inner query needs to be nested under a `"match"` field, as per: https://www.elastic.co/guide/en/elasticsearch/reference/6.0/query-dsl-span-first-query.html


sksamuel:
Excellent thank you.

PR #1123, opened 2017-10-18
Fix SpanMultiTermQueryBodyFn

catalin-ursachi:
Analogous fix to https://github.com/sksamuel/elastic4s/pull/1121, in accordance with https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-span-multi-term-query.html 🙂


catalin-ursachi:
The failure seems spurious, as far as I can tell; it passes on my computer on both master and the feature branch.

Also, as far as I can tell, the other span queries do seem to match the API.

sksamuel:
:+1:

Issue #1039, opened 2017-08-23
HttpClient does not URL encode document IDs


kbedel:
The HttpClient does not URL encode document IDs. This causes issues when trying to do things like index/get/delete documents where the ID has characters such as a space in it. This IDs work fine in the TcpClient.

For example, this test passes:
```
"tcp client" should {
  "index a document" when {
    "id has a space" in {
      client.execute {
        indexInto("test" / "indextest").doc("{}").id("test id")
      }.await
    }
  }
}
```

while this one:
```
"http client" should {
  "index a document" when {
    "id has a space" in {
      httpClient.execute {
        indexInto("test" / "indextest").doc("{}").id("test id")
      }.await
    }
  }
}
```

results in the following error:
```
Illegal character in path at index 20: /test/indextest/test id
java.lang.IllegalArgumentException: Illegal character in path at index 20: /test/indextest/test id
	at org.elasticsearch.client.RestClient.buildUri(RestClient.java:576)
	at org.elasticsearch.client.RestClient.performRequestAsync(RestClient.java:320)
	at org.elasticsearch.client.RestClient.performRequestAsync(RestClient.java:268)
	at com.sksamuel.elastic4s.http.HttpExecutable$RichRestClient.$anonfun$async$5(HttpExecutable.scala:49)
	at com.sksamuel.elastic4s.http.HttpExecutable$RichRestClient.$anonfun$async$5$adapted(HttpExecutable.scala:49)
	at com.sksamuel.elastic4s.http.HttpExecutable$RichRestClient$$Lambda$1552/1889365477.apply(Unknown Source)
	at com.sksamuel.elastic4s.http.HttpExecutable$RichRestClient.future(HttpExecutable.scala:26)
	at com.sksamuel.elastic4s.http.HttpExecutable$RichRestClient.async(HttpExecutable.scala:50)
	at com.sksamuel.elastic4s.http.index.IndexImplicits$IndexHttpExecutable$.execute(IndexImplicits.scala:39)
	at com.sksamuel.elastic4s.http.index.IndexImplicits$IndexHttpExecutable$.execute(IndexImplicits.scala:12)
	at com.sksamuel.elastic4s.http.HttpClient.execute(HttpClient.scala:23)
	at com.sksamuel.elastic4s.http.HttpClient.execute$(HttpClient.scala:23)
	at com.sksamuel.elastic4s.http.HttpClient$$anon$1.execute(HttpClient.scala:37)
```

[...]

PR #1129, opened 2017-11-04
HttpClient does not URL encode document IDs #1039

francisdb:
1b1065b01d09d39c5a7b797a35acf169a20bd2aa + some other related fixes applied to 5.6.x
HttpClient does not URL encode document IDs #1039


sksamuel:
:+1:

PR #1133, opened 2017-11-13
support versioning in bulk index

yanns:
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-versioning

### DIFF
--- a/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/bulk/BulkBuilderFn.scala
+++ b/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/bulk/BulkBuilderFn.scala
@@ -2,6 +2,7 @@ package com.sksamuel.elastic4s.http.bulk
 
 import com.sksamuel.elastic4s.bulk.BulkDefinition
 import com.sksamuel.elastic4s.delete.DeleteByIdDefinition
+import com.sksamuel.elastic4s.http.index.VersionTypeHttpString
 import com.sksamuel.elastic4s.http.update.UpdateBuilderFn
 import com.sksamuel.elastic4s.indexes.{IndexContentBuilder, IndexDefinition}
 import com.sksamuel.elastic4s.json.XContentFactory
@@ -21,6 +22,8 @@ object BulkBuilderFn {
         index.id.foreach(id => builder.field("_id", id.toString))
         index.parent.foreach(builder.field("_parent", _))
         index.routing.foreach(builder.field("_routing", _))
+        index.version.foreach(builder.field("_version", _))
+        index.versionType.foreach(versionType ⇒ builder.field("_version_type", VersionTypeHttpString(versionType)))
         builder.endObject()
         builder.endObject()
 
@@ -50,6 +53,8 @@ object BulkBuilderFn {
         builder.field("_id", update.id)
         update.parent.foreach(builder.field("_parent", _))
         update.routing.foreach(builder.field("_routing", _))
+        update.version.foreach(builder.field("_version", _))
+        update.versionType.foreach(builder.field("_version_type", _))
         builder.endObject()
         builder.endObject()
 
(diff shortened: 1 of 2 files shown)


sksamuel:
:+1:

yanns:
@sksamuel do you have a rough idea when you will push a new version?
(just asking - no pressure)

sksamuel:
This week for the 6.0 release. Before thanksgiving on Thurs.

yanns:
@sksamuel sorry for up-ing this but I don't see any new version in the meantime. Can we help you somehow for that?

sksamuel:
Yes delayed because I am struggling to decide on this:

In an API, when dealing with Future's, for a case where you have three types of things:

1. The success case
2. The system dealt with your request but returned an error, like say "invalid search term" or something
3. An error like host not found, connection timeout etc.

Do you think it's better to have:

`Future[Either[RequestSuccess, RequestFailure]]` and then have case 3 above as the `Exception` case of a Future.
This means you have to pattern match on the success case.

Or

`Future[RequestSuccess]` and then model case 2 above in the form of an Exception which is a subtype of `Exception` (case 3). This means when handling errors you pattern match on the error case.

Or

Third way.

yanns:
@I'll discuss that with @jadlr who is also using the lib quite intensively, and we get back to you.
Do you want to discuss that on this PR, or do you prefer that we open an issue for that?

sksamuel:
Can do it on email, or an issue or here I don't mind. It's really the last thing I want to get done before I release, and will only take a day once I decide. I've just been flip flopping between ideas, and I want to get it right before I cement in place 6.0. Especially as this is the time the rest of TCP users will finally move to the HTTP api.

yanns:
@sksamuel With @jadlr, we'd prefer the following approach:
`Future[Either[RequestFailure, RequestSuccess]]` (failure on the left, success on the right)

- `Future[Right[A]]` is the success case (ex: elasticsearch answers with 200, 201...) (the 1st case)
- `Future[Left[RequestFailure]]` is the expected failure case (ex: elasticsearch answers with 400)  (the 2nd case)
- `Future.failure` is the unexpected failure (connection problem, 503 errors).  (the 3rd case)

In both failure cases, it's important that the client has all possible information to be able to react.
For ex, we can retry with a backoff in case of 503.

sksamuel:
What I did last night and is in master (but happy to change it round again) was this:

`Future[Response[U]]` where Response is a sealed trait of `RequestSuccess[U]` and `RequestFailure` and `Future.failure` is the unexpected failure.

The `Response` object comes with status code, full json body, all http headers.

yanns:
I look at master. At first sight, I think you forgot to make `Response` a sealed trait (it's a normal trait)

Some arguments to favor `Either` instead of the sealed trait:
- we have more and more tools to work with `Either` in the scala community. For example, cats have an Either transformer that makes it easy to compose different `Future[Either[E, _]]`.
- In the `Response` you are implementing `map`, `fold`, `foreach` that are already implemented in `Either`:
```
val e: Either[String, Int] = Right(42)
e.fold(identity, _.toString)

import cats.syntax.either._
e.map(_ + 2)
e.foreach(println)
```

But the sealed trait is not a big issue. We can still transform it to an `Either` if needed.
Actually, if you want to go with the sealed trait, you could add:
```
sealed trait Response[+U] {
  def toEither: Either[RequestFailure, RequestSuccess[U]]
}
```

sksamuel:
Good feedback. My rationale for using a sealed trait (I will seal it :)) instead of an Either is that I was sick of typing out Either[RequestFailure, X] everywhere :)
I could just make a type alias. Maybe I'll change it back to Either now and then get this release out.

sksamuel:
If HttpClient returns `Future[Either[RequestFailure, RequestSuccess[U]]]` then I think we get the best of all worlds. An either for easy interop with existing scala tools, and the types that contain status, body, and headers.

yanns:
When I use a lot of `Either[E, _]` where `E` is fixed, I usually introduce a type alias.

Another notice:
```
def get: U // returns the marshalled response U or throws an exception
```

Maybe you could add a method returning an Either[ParseError, U], and makes `get`using it:
```
def deserialize: Either[ParseError, U]
def get: U = deserialize.fold(e ⇒ throw new Exception(e), identity)
```
So that clients can decide which level of "FP purines" they want.

yanns:
> If HttpClient returns Future[Either[RequestFailure, RequestSuccess[U]]] then I think we get the best of all worlds. An either for easy interop with existing scala tools, and the types that contain status, body, and headers.

👍

sksamuel:
It's done, I just need to update the tests. Then I will do the 6.0 release.

yanns:
you're awesome, thx!!!

jadlr:
Thanks @sksamuel! Looking forward to the release

PR #1138, opened 2017-11-17
Make InnerHits recursive

douglaz:
Yes, that happens and it's actually useful.


sksamuel:
A test would be a good addition as well.

douglaz:
> A test would be a good addition as well.
I agree, but I won't able to look into that until right know. I've tested with real data and it works.

PR #1144, opened 2017-11-27
Make it possible to make http request with http entity from InputStream/File

chikei:
The idea is that in memory constrained/sending large document environment we could provide custom HttpExecutable which provide an InputStream generated json on the fly or write the content to file to avoid memory overhead due to building json string for current string based http entity


sksamuel:
Great thanks.

PR #1150, opened 2017-11-29
Fix negative boost of fields by default (fixes #1148)

### LINKED ISSUE #1148: Getting silly results with field^-1
- elastic4s 5.6.0
- elasticsearch 6.0
- scala 2.12.4

When calling `fields("foo")` without specifying a boost, elastic4s sends `foo^-1.0` to elasticsearch. Instead of omitting the boost by just sending `foo`, it adds a default boost of `-1.0`.

Maybe that worked well before elasticsearch 6.0, I don't know. What I know is that with elasticsearch 6.0, `foo` and `foo^-1.0` return very different results. It seems that the query term will actually be avoided when the boost is `-1.0`. Which kind of makes sense.

Here's my query generated by elastic4s:
```json
                "multi_match":{
                  "query":"dragon",
                  "fields":[
                      "chapterNames^-1.0",
                      "chapterTexts^-1.0",
                      "members^-1.0",
                      "name^-1.0",
                      "owner^-1.0"
                  ],
                  "type":"most_fields",
                  "operator":"OR",
                  "analyzer":"english",
                  "slop":0,
                  "prefix_length":0,
                  "max_expansions":50,
                  "lenient":false
                  "zero_terms_query":"NONE",
                  "boost":1.0
                }
```

It returns results that never contain the keyword "dragon". When I fix the query by removing the `^-1.0`, like this:

```json
                "multi_match":{
                  "query":"dragon",
                  "fields":[
                      "chapterNames",
                      "chapterTexts",
                      "members",
                      "name",
                      "owner"
                  ],
                  "type":"most_fields",
                  "operator":"OR",
                  "analyzer":"english",
                  "slop":0,
                  "prefix_length":0,
                  "max_expansions":50,
                  "lenient":false
                  "zero_terms_query":"NONE",
                  "boost":1.0
                }
```
then I get correct results containing the "dragon" keyword.

avdv:


elastic4s-tcp/src/main/scala/com/sksamuel/elastic4s/searches/queries/QueryStringBuilderFn.scala
```diff
@@ -15,7 +15,7 @@ object QueryStringBuilderFn {
     query.defaultField.foreach(builder.defaultField)
     query.enablePositionIncrements.foreach(builder.enablePositionIncrements)
     query.fields.foreach {
-      case (name, -1) => builder.field(name)
+      case (name, 1D) => builder.field(name)
       case (name, boost) => builder.field(name, boost.toFloat)
```
avdv: Note, this seems to be a subtle bug by itself since it maches on an Int and never reaches the first case.

avdv:
closing since this should be fixed differently

sksamuel:
Did you mean to reopen this or am I ok to close ?

avdv:
No, that was unintentional. /close

PR #1163, opened 2017-12-12
declare cats dependency in community build friendly way

SethTisue:
this is needed to keep dbuild from getting confused (context: https://github.com/scala/community-builds/pull/649)

...but, it also seems more correct to me to declare the slimmer dependency.

this PR is against the 5.4.x branch; the same change
already exists on master (in 70833ba096d34c)


sksamuel:
You don't need a release do you because you're building from source.

SethTisue:
correct

PR #1167, opened 2017-12-12
Remove ExecutionContext.Implicits.global from main library code

solicode:
After upgrading to 6.0.4, I noticed lots of strange issues in one of my Play apps with futures (timeouts, thread starvation, etc). It looks like the issue is coming from the use of `scala.concurrent.ExecutionContext.Implicits.global` in `HttpClient` (Play/Akka uses its own ExecutionContext, and it's problematic to add Implicits.global into the mix).

This PR is for addressing that by making the ExecutionContext an implicit parameter so that the caller has the option to specify it.

Also, I noticed `HttpClient.execute` is using a Promise which seems unnecessary since we're already working with a Future. I updated that too to get rid of the unnecessary overhead.

Unfortunately I had to touch pretty much all the tests too. I just used `scala.concurrent.ExecutionContext.Implicits.global` there as it doesn't matter for the tests.

### DIFF
--- a/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/HttpClient.scala
+++ b/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/HttpClient.scala
@@ -9,8 +9,7 @@ import org.apache.http.HttpHost
 import org.elasticsearch.client.RestClient
 import org.elasticsearch.client.RestClientBuilder.{HttpClientConfigCallback, RequestConfigCallback}
 
-import scala.concurrent.{Future, Promise}
-import scala.util.{Failure, Success}
+import scala.concurrent.{ExecutionContext, Future}
 
 sealed trait Response[+U] {
   def status: Int // the http status code of the response
@@ -43,8 +42,6 @@ case class RequestFailure(override val status: Int, // the http status code of t
 
 trait HttpClient extends Logging {
 
-  import scala.concurrent.ExecutionContext.Implicits.global
-
   // the underlying client that performs the requests
   def client: HttpRequestClient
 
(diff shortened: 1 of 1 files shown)


elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/HttpClient.scala
```diff
@@ -59,24 +56,14 @@ trait HttpClient extends Logging {
...
-            p.trySuccess(Left(failure))
-        } catch {
-          case t: Throwable => p.tryFailure(t)
-        }
-      case Failure(t) => p.tryFailure(t)
-    }
-    p.future
+  def execute[T, U](request: T)(implicit exec: HttpExecutable[T, U], executionContext: ExecutionContext): Future[Either[RequestFailure, RequestSuccess[U]]] = {
+    exec.execute(client, request)
+      .map(r => exec.responseHandler.handle(r) match {
```
sksamuel: I don't think this handles the case where the response handler throws an exception, which I'd like to see materialized as a Future.failure like before.

sksamuel:
In general this change is probably required, but I'm relutant to bring it into 6.0.x and break everyone's code as they'll now need the implicit in scope. What about making it the default for the implicit and in 6.1 we drop the default?

elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/HttpClient.scala
solicode: It wouldn't be safe in `onComplete`, but I don't think exceptions in `map` escape the context of Future.

```scala
val f = Future.successful("not a number").map(_.toInt)
f.failed.map(_ => println("failed future"))
```

results in a proper Future.failed.

https://docs.scala-lang.org/overviews/core/futures.html

> If the mapping function throws an exception the future is completed with that exception.

I think the 2 pieces of code are functionally equivalent, unless I'm forgetting to take something into account.

solicode:
Sure, the default parameter idea sounds reasonable for 6.0.x. I updated the code.

elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/HttpClient.scala
sksamuel: Ok

sksamuel:
👍

PR #1168, opened 2017-12-13
 Bucket selector pipeline aggregation support in elastic4s-http

lmafonso:
Hello,
Could you please review these changes?
Tks


sksamuel:
Looks good, but will need to fix the conflict so I can merge.

lmafonso:
Conflict fixed. 
However, this code differs from recent changes, since other builder objects have been placed directly into AggregationBuilderFn.scala, while I was trying to reproduce a similar package structure from the tcp client.
Let me know if I should move BucketSelectorPipelineBuilder to the same file, to make it consistent with the other changes. Tks.

sksamuel:
👍

PR #1175, opened 2017-12-16
Range aggregation support (results + keyed)

TrustNoOne:
I added support for keyed and unkeyed range aggregation results and keyed date rage aggregations


sksamuel:
Hi. Wasn't key already available inside the date range class ?

TrustNoOne:
I copied the structure from filters/keyedFilters, key is either inside the bucket or as an actual key in the map of buckets. 

This could be changed in a number of ways:
- the key could be copied in the bucket object even in case of keyed results, making it non-optional. 
- we could always build a map from the results (keyed or unkeyed) and make the Seq[Buckets] something like `lazy val buckets: Seq[RangeBucket] = bucketsMap.values.toSeq`

What do you think?

sksamuel:
👍

PR #1181, opened 2017-12-18
Added explicit 'not implemented' errors to aggregation and query builder functions

catalin-ursachi:
I think an explicit exception explaining that a given query or aggregation hasn't been implemented should be more user friendly than a `MatchError` (which might make the user think they've done something wrong, at first).


sksamuel:
Good idea.

PR #1182, opened 2017-12-18
Added terms lookup query support in http client

TrustNoOne:
Hi,
terms lookup query is broken in the http client.
I see there are two ways to specify it, a `ref` inside a `TermsQueryDefinition` or directly a `TermsLookupQueryDefinition`. 
I implemented both: the former was bugged (wrongly nested json) and the latter was missing in `QueryBuilderFn` (MatchError)


sksamuel:
Excellent.

PR #1185, opened 2017-12-18
Add PrintableRequest trait; define SearchPrintableRequest

### LINKED ISSUE #1180: Shows could print out path as well as Json body
At the moment, using `httpClient.show` only provides partial info about the request, as it doesn't print out the URL path. Another consequence of that is that `Show`s are only defined for requests that do carry Json, so a method taking a generic request can't always take a `Show` as an evidence parameter.

Would it be worthwhile changing the `Show`s to render the request path as well as the Json body, and implementing them for non-Json requests as well?

catalin-ursachi:
Fixes #1180

### DIFF
--- a/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/PrintableRequest.scala
+++ b/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/PrintableRequest.scala
@@ -1,0 +1,18 @@
+package com.sksamuel.elastic4s.http
+
+import com.sksamuel.exts.Logging
+
+/**
+  * @tparam T the type of the request object to be printed
+  */
+abstract class PrintableRequest[T] extends Logging {
+
+  def endpoint(request: T): String
+
+  def body(request: T): Option[String] = None
+
+  /**
+   * Print the endpoint and JSON body (if applicable) of the given request.
+   */
+  def print(request: T): String = (endpoint(request) +: body(request).toSeq).mkString("\n")
+}
--- a/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/search/SearchImplicits.scala
+++ b/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/search/SearchImplicits.scala
@@ -3,7 +3,7 @@ package com.sksamuel.elastic4s.http.search
 import java.net.URLEncoder
 
 import cats.Show
-import com.sksamuel.elastic4s.http.{HttpEntity, HttpExecutable, HttpRequestClient, HttpResponse, IndicesOptionsParams, ResponseHandler}
+import com.sksamuel.elastic4s.http._
 import com.sksamuel.elastic4s.json.JacksonSupport
 import com.sksamuel.elastic4s.searches.queries.term.{BuildableTermsQuery, TermsQueryDefinition}
 import com.sksamuel.elastic4s.searches.{MultiSearchDefinition, SearchDefinition, SearchType}
@@ -20,7 +20,13 @@ trait SearchImplicits {
   }
 
   implicit object SearchShow extends Show[SearchDefinition] {
-    override def show(req: SearchDefinition): String = SearchBodyBuilderFn(req).string()
+    override def show(req: SearchDefinition): String = SearchHttpExecutable.body(req)
+  }
+
+  implicit object SearchPrintableRequest extends PrintableRequest[SearchDefinition] {
+    override def body(req: SearchDefinition): Option[String] = Some(SearchHttpExecutable.body(req))
+
+    override def endpoint(req: SearchDefinition): String = SearchHttpExecutable.endpoint(req)
   }
 
   implicit object MultiSearchShow extends Show[MultiSearchDefinition] {
@@ -62,15 +68,6 @@ trait SearchImplicits {
 
     override def execute(client: HttpRequestClient, request: SearchDefinition): Future[HttpResponse] = {
 
-      val endpoint = if (request.indexesTypes.indexes.isEmpty && request.indexesTypes.types.isEmpty)
-        "/_search"
-      else if (request.indexesTypes.indexes.isEmpty)
-        "/_all/" + request.indexesTypes.types.map(URLEncoder.encode).mkString(",") + "/_search"
-      else if (request.indexesTypes.types.isEmpty)
-        "/" + request.indexesTypes.indexes.map(URLEncoder.encode).mkString(",") + "/_search"
-      else
-        "/" + request.indexesTypes.indexes.map(URLEncoder.encode).mkString(",") + "/" + request.indexesTypes.types.map(URLEncoder.encode).mkString(",") + "/_search"
-
       val params = scala.collection.mutable.Map.empty[String, String]
       request.requestCache.map(_.toString).foreach(params.put("request_cache", _))
       request.searchType.filter(_ != SearchType.DEFAULT).map(SearchTypeHttpParameters.convert).foreach(params.put("search_type", _))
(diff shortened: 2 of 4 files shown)


elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/search/SearchImplicits.scala
```diff
@@ -20,7 +20,13 @@ trait SearchImplicits {
   }
 
   implicit object SearchShow extends Show[SearchDefinition] {
-    override def show(req: SearchDefinition): String = SearchBodyBuilderFn(req).string()
+    override def show(req: SearchDefinition): String = SearchHttpExecutable.body(req)
+  }
+
+  implicit object SearchPrintableRequest extends PrintableRequest[SearchDefinition] {
```
catalin-ursachi: Worth considering whether it would be better to just have `SearchHttpExecutable` implement `PrintableRequest[SearchDefinition]` (and indeed `Show[SearchDefinition]`), rather than having them be defined separately?

sksamuel: I think maybe separate objects for now so each implementation isn't massive ?

elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/PrintableRequest.scala
```diff
@@ -0,0 +1,18 @@
+package com.sksamuel.elastic4s.http
+
+import com.sksamuel.exts.Logging
+
+/**
+  * @tparam T the type of the request object to be printed
+  */
+abstract class PrintableRequest[T] extends Logging {
+
+  def endpoint(request: T): String
```
sksamuel: What do you think about having method / params defined as seperate methods? So people don't have to parse out parameters should they wish to.

sksamuel: What do you think about having params and method defined seperately so people don't have to parse out params if they want to use them.

elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/search/SearchImplicits.scala
sksamuel: Although if I guess there's duplication now otherwise, so yeah maybe merge them in.

elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/PrintableRequest.scala
catalin-ursachi: Ah, you're right; I'd forgot about the method, and separate parameters makes sense!

sksamuel: So we need endpoint, method, params, and body.

sksamuel: And maybe one that brings together method /endpoint?params

catalin-ursachi: Should all of these be exposed on `implicit class RichRequest[T]`?

sksamuel: What about introducing something like `Request` which brings together everything needed to model the request - body, params, method, endpoint. Then the `HttpExecutable` can become `Executable` which simply takes the `Request` as a parameter and `PrintableRequest` can also accept that `request` instance? I'm just brainstorming, might be a rubbish idea.

sksamuel: You might need to move stuff around to get it to work. Would printable request then just be another instance of Show, but instead of `Show[IndexDefinition]` for example, it could be `Show[Request[T]]` which could print out body + header. And if people just want the header + body separately, they've got the `Request` itself instead.

sksamuel: If this seems like it will be a good change, I'm happy for you to refactor just a single request, and push up a broken build, otherwise you would need to change every single request before we can review it.

catalin-ursachi: Hmm; could you clarify what you had in mind here? Would `Request[T]` be a static object, akin to `HttpExecutable[T, R]`, which takes the actual `T` instance as a parameter on its methods? If so, then we couldn't have a `Show[Request[IndexDefinition]]`, as it wouldn't have the actual request data.

catalin-ursachi: Or, should `Request[T]` be an implicit class, wrapping a `T`?

sksamuel: I can have a little play myself if that's ok, see what I can come up with.

sksamuel: If we introduce a class that models a http request:

`case class Request[T](method: Method, endpoint: String, params: Map[String, String], body: Option[String])
`

Then we the existing `HttpExecutable` typeclass that takes an elasticsearch operation and then executes it can be split up into a typeclass that builds the `Request` object from the operation class.

```scala
trait RequestBuilder[T] {
  def build(t: T): Request[T]
}
```

Eg for search.

```scala
  implicit object SearchRequest extends RequestBuilder[SearchDefinition] {

    override def build(req: SearchDefinition): Request[SearchDefinition] = {

      val endpoint: String = {

        def encodedIndexes = req.indexesTypes.types.map(URLEncoder.encode(_, "UTF8")).mkString(",")

        if (req.indexesTypes.indexes.isEmpty && req.indexesTypes.types.isEmpty) "/_search"
        else if (req.indexesTypes.indexes.isEmpty) "/_all/" + encodedIndexes + "/_search"
        else if (req.indexesTypes.types.isEmpty) "/" + encodedIndexes + "/_search"
        else "/" + encodedIndexes + "/" + req.indexesTypes.types.map(URLEncoder.encode(_, "UTF8")).mkString(",") + "/_search"
      }

      val params: Map[String, String] = {
        val params = scala.collection.mutable.Map.empty[String, String]
        req.requestCache.map(_.toString).foreach(params.put("request_cache", _))
        req.searchType.filter(_ != SearchType.DEFAULT).map(SearchTypeHttpParameters.convert).foreach(params.put("search_type", _))
        req.control.routing.map(_.toString).foreach(params.put("routing", _))
        req.keepAlive.foreach(params.put("scroll", _))

        req.indicesOptions.foreach { opts =>
          IndicesOptionsParams(opts).foreach { case (key, value) => params.put(key, value) }
        }
        params.toMap
      }

      Request[SearchDefinition](Post, endpoint, params, SearchBodyBuilderFn(req).string().some)
    }
  }
```

Now the public API has a way of getting access to the header information. This could then be further codified in a `Show` typeclass.

```scala
object Request {
  implicit def requestShow[T]: Show[Request[T]] = new Show[Request[T]] {
    override def show(t: Request[T]): String = t.method + " " + t.endpoint + "?" + t.params + "\n" + t.body.getOrElse("")
  }
}
```

So then if we want to print out the search request, we can either get a handle to the request object and print out the bits we want:

```scala
val op = search("index").termQuery("name", "bob")

val req = SearchRequest.build(op)
// or if we add some implicit syntax
val req = op.toRequest

println(req.params)
```

Or we can use the show typeclass.

```scala
val op = search("index").termQuery("name", "bob")
// with some implicit syntax
println(op.toRequest.show)
```

catalin-ursachi:
Ah, sorry for the delay; I'll try to get this work done this weekend. 🙂

catalin-ursachi:
Hi @sksamuel; sorry it took a while. I've created https://github.com/sksamuel/elastic4s/pull/1257 to track the changes requested in that latest comment; I think it makes sense to keep them separate from the rest of this.

PR #1186, opened 2017-12-18
ExtendedBounds now supports both dates and numbers

TrustNoOne:
Hi, 
This is my try at "extending ExtendedBounds". 

Let me know what you think.
I haven't added it to HistogramAggregation but it should be pretty straight forward 
`bounds: NumericExtendedBounds` 

Feel free to modify it or not use it altogether if you don't like it


sksamuel:
Excellent work.

PR #1187, opened 2017-12-18
Refactored geo shape query model to support multiple shapes

runebarikmo:
Hi,

I tried to use the reworked geoshape query builder in the http module, but could not get it to work. I think the the main problem is that InlineShape is modeled as a sequence of points when in fact not all shapes looks like that. E.g. Circle has a different shape from Polygon etc.

I've made an attempt to fix this by refactoring the geo shape query builder a bit. All of the shapes defined in this document should now be supported:

https://www.elastic.co/guide/en/elasticsearch/reference/current/geo-shape.html

Changes:
- Add new ADTs to represent the single geojson shapes and  GeometryCollection which is a collection of geojson shapes
- Refactor GeoShapeQueryBodyFn query builder to support the new geo shape model
- Change GeoShapeQueryDefinition to work with the new model
- Add query builder test for each query

What do you think? I'm happy to make adjustments if you have suggestions.

### DIFF
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/json/XContentBuilder.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/json/XContentBuilder.scala
@@ -33,6 +33,35 @@ class XContentBuilder(root: JsonNode) {
     this
   }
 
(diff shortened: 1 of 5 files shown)


sksamuel:
Excellent stuff.

PR #1192, opened 2017-12-21
1188-fix-address-aws-normalized-uri-encoding

### LINKED ISSUE #1188: AWS request signature does not match on getAliases
This might be me just misunderstanding how i should use the `Aws4ElasticClient`, but here goes.
This is with elasticsearch 6.0 in aws and elastic4s 6.0.1.

`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY` and `AWS_DEFAULT_REGION` is set via environment variables.
Then i try to do:
```
val endpoint = "elasticsearch://awsendpointurl.com:443?ssl=true"
val client = Aws4ElasticClient(ep)

client.execute {
  getAliases("images", Nil)
}
```

And this works fine when requesting from a local elasticsearch container (which is expected since it does not require signing the request)

However, when attempting to do this towards aws, the Either returns left with a message of: 
```
The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details.

The Canonical String for this request should have been
'GET
/images/_alias/

host:awsendpointurl.com
user-agent:Apache-HttpAsyncClient/4.1.3 (Java/1.8.0_144)
x-amz-date:20171218T142308Z

host;user-agent;x-amz-date
somelongstringhere'

The String-to-Sign should have been
'AWS4-HMAC-SHA256
20171218T142308Z
20171218/eu-central-1/es/aws4_request
adifferentlongstringhere'
```

This seems to happen with `getAliases`, but not with `createMapping`, which i find weird when the same command works locally.

saraiva132:
Fixes #1188 

Not surprisingly java URLEncoder does not respect RFC 3986. Had to some fixes and added a test that covers url forbidden characters.

### DIFF
--- a/elastic4s-aws/src/main/scala/com/sksamuel/elastic4s/aws/CanonicalRequest.scala
+++ b/elastic4s-aws/src/main/scala/com/sksamuel/elastic4s/aws/CanonicalRequest.scala
@@ -1,13 +1,11 @@
 package com.sksamuel.elastic4s.aws
 
 import com.sksamuel.elastic4s.aws.Crypto._
+import java.net.{URI, URLEncoder}
 
-import java.net.{ URI, URLEncoder }
-import java.nio.charset.Charset
-
-import org.apache.http.{ HttpEntityEnclosingRequest, HttpRequest }
+import org.apache.http.{HttpEntityEnclosingRequest, HttpRequest}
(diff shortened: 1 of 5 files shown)


sksamuel:
👍

PR #1193, opened 2017-12-21
Fix: support update of sequences and nested fields

runebarikmo:
Update of nested fields and sequences seemed to have stopped working due to conversions between scala and java collection which made pattern matching on the field type fail. I've made a minor change to fix this problem,


sksamuel:
👍

PR #1194, opened 2017-12-22
Fix missing otherBucket and otherBucketKey in KeyedFilters request

kgston:
The options were available in the definition but were never written into the request to ES.

Added tests for this issue


sksamuel:
:+1:

PR #1205, opened 2017-12-29
 Fix Extended Stats Aggregation in http client (test included)

lmafonso:
Could you review, please? Thanks!


sksamuel:
awesome :+1:

PR #1206, opened 2017-12-30
Parameterize over effect type.

daviswahl:
The goal of this commit is to allow the execute method of elastic4s to operate
over arbitrary effect types, eg: instead of returning a Scala Future, we can easily return
a Twitter Future, cats IO, or scalaz Task. The benefit of this should be pretty obvious: Manyusers (myself, for instance) don't use Scala Futures, and having to import an execution context and dispatch unnecessary work into a thread pool, while not critically important, is pretty annoying.

The downside is that users will need to include an extra import (to pull in the correct
executor), along with adding extra complexity to the code base.

Notes on the PR:

1. This currently compiles and works. The tests module fails to compile, but I'm fairly confident it will work once the implicits are available. I wanted to get this in front of you before doing all of that work though, because there are a ton of tests!

2. The files you want to focus on are `FromListener`, a typeclass that describes a type that can convert from a `ResponseListener`, along with `HttpClient`, `HttpExecutable` and `ElasticSearchJavaRestClient`, which are changed to operate over the FromListener typeclass. Everything else is just find and replace, for the most part.

3. Ideally, I think we'd parameterize the effect on the client rather than on the async/execute methods, but it wasn't immediately obvious how this can be done without a really aggressive refactor.

4. The only thing that I think really needs a closer look is the SearchScroll api, because it uses Await to wait for each step. I added an `Awaitable` typeclass to deal with this but I dont think it's ideal, and it's totally untested.

5. While I'm confident that this more or less works, it's rough, needs documentation, needs tests, and needs a lot of cleanup. I'm making this PR to see if you're interested in it before I do all the extra work. This surprisingly only took a few hours, so if you're not interested in taking things in this direction, I won't be angry.

Let me know if you have any questions!


sksamuel:
This looks really interesting!

I haven't read through this PR yet, but before I do, I'll say I don't know much about effects. However I do believe in keeping the API simple for people who are not on the high end of FP (like myself). I wouldn't want to merge anything that raises the bar for users who are not _expert_ Scala people, as that's where I think some Scala libraries go wrong.

sksamuel:
I've had a look through now, some great looking code, but I have some questions as it's at the limit of my knowledge of cats.

What does `FromListener` do - is that the typeclass that will convert from the http client listener into an `F[T]` ? So if you wanted to use cats IO you'd need a `FromListener[IO]` or whatever? 

The requests are being executed on the apache async http client's thread pool, so by using a Task or IO are you actually deferring the execution, or just basically wrapping that? If you are just wrapping, is it easier to just wrap the resultant Future?

daviswahl:
> I haven't read through this PR yet, but before I do, I'll say I don't know much about effects.

Don't worry, that makes two of us. I'm probably misusing the term `Effect` here, which probably means something really specific in the FP world. I just mean "the category of things like Future, IO, Task, etc." 

> I wouldn't want to merge anything that raises the bar for users who are not expert Scala people, as that's where I think some Scala libraries go wrong.

Totally agree: As I think you've figured out, all that would change is that the user would need to import an instance of `FromListener` for their desired effect type -- basically the same way you handle the various json libraries currently. I'm hoping there's a way to do this that will allow the library to default to the Scala Future effect, otherwise this will constitute a breaking change and require probably waiting for the next major release.

> What does FromListener do - is that the typeclass that will convert from the http client listener into an F[T] ? So if you wanted to use cats IO you'd need a FromListener[IO] or whatever?

Yep, you've got it. I've got some samples of what this would look like that I can post in a bit.

> The requests are being executed on the apache async http client's thread pool, so by using a Task or IO are you actually deferring the execution, or just basically wrapping that? If you are just wrapping, is it easier to just wrap the resultant Future?

Great question. In my use case, I'm actually converting to a Twitter Future, so the underlying semantics are similar: My main goal is to avoid the overhead of wrapping the future. As I mentioned, I really doubt the impact here is huge, but it is something. I'll get some quantifiable measurements here and get back to you.

However, for cases like IO, I believe you're correct: The impact would be that execution is entirely deferred until "the end of the world," which is desirable if you're operating in the cats world. I'm not an expert on IO/Effect though, so I'd need to test this. I have a working instance of `FromListener` for IO: I'll take it for a test run and let you know how it behaves.

Oh and happy new year!

daviswahl:
Quick update:

1. I added a default instance of FromListener for Future, so we no longer require any new imports. Users who wish to continue using scala futures should not need to make any changes. This has the nice effect that the tests all seem to pass, *I think*. I'm having issues running the tests locally, (too many open files, which happens frequently on my macbook). Travis seems happy though.

2. Demonstrated the behavior of using this with the IO effect. As we discussed, it looks like it does defer execution until "the end of the world." I did the work in another branch so as to not pollute this PR, you can see the changes here: https://github.com/daviswahl/elastic4s/commit/17c9bcbc7dc2dc67cfdaaf2f4b821ff5e2d6c2c0

3. Still need to do cleanup and write some extra tests. Also this is a big change and I don't expect you to just merge it now that tests are passing.

sksamuel:
It's looking good. I'd want to change a few things to put my spin on it, as I'll be maintaining this of course once it's merged. 

I don't think the FromListener will work (correct me if I am mistaken), as you pass in a callback, so at that point you've already begun execution - thus a cats `IO` is just wrapping the already in progress work. Perhaps the typeclass really needs to be something like `Executable` which takes the client and the request, and then returns the effect. 

Also take a look at my other work in this PR (at the bottom) which is a partial rework of the HttpExecutable to better separate request's from the executions. It might be something that would help in this. https://github.com/sksamuel/elastic4s/pull/1185

sksamuel:
Also, I think we could ship a scala Future version by default, a cats IO version available in core, and then an elastic4s-scalaz package which offers a Task version, and an elastic4s-twitter which offers a twitter Future version (if you think there's demand, I didn't realize twitter Futures were still in use).

daviswahl:
> I don't think the FromListener will work (correct me if I am mistaken), as you pass in a callback, so at that point you've already begun execution - thus a cats IO is just wrapping the already in progress work.

I think you're partially correct: this is true (and the correct behavior) in the case of the Future instance(s), but not true of Effect / IO. Here's the implementation for effect:

```scala
  implicit def FromListenerForEffect[F[_]: Effect]: FromListener[F] = new FromListener[F] {
    override def fromListener(f: ResponseListener => Unit): F[HttpResponse] = {
      Effect[F].async { callback =>
        f(new ResponseListener {
          override def onSuccess(r: org.elasticsearch.client.Response): Unit =
            callback(Right(fromResponse(r)))

          override def onFailure(e: Exception): Unit = e match {
            case re: ResponseException => callback(Right(fromResponse(re.getResponse)))
            case t                     => callback(Left(JavaClientExceptionWrapper(t)))
          }
        })
      }
    }
```

In this case, `f` is the method that creates our `ResponseListener` and starts the work, and `f` isn't called until `async` is called (which isn't until "the end of the world").  The test in https://github.com/daviswahl/elastic4s/commit/17c9bcbc7dc2dc67cfdaaf2f4b821ff5e2d6c2c0 pretty *effect*ively demonstrates that this is the case, I think. I also did some fairly basic printf testing as well and am satisfied that things are working the right way. I'll be honest though: this is where my understanding of things starts to break down, and reading the Effect code certainly doesn't help my understanding, so I'll happily provide some extra tests. 

Either way, your suggestion about `Executable` may be worthwhile. I'll take a look at that PR and do some experimentation in that direction and get back to you.  

> Also, I think we could ship a scala Future version by default, a cats IO version available in core, and then an elastic4s-scalaz package which offers a Task version, and an elastic4s-twitter which offers a twitter Future version (if you think there's demand, I didn't realize twitter Futures were still in use). Edit: Maybe a monix version too.

Sounds great. IO would require pulling in the cats-effect package, so it might still be worth breaking that into a separate package (elastic4s-cats-effect). 

Also: We wouldn't be offering a `FromListener` instance for `IO`, we'd be offering an instance for `Effect` (for which an `IO` instance exists). One nice thing about this is that we'd be able to offer instances of FromListener for anything with an Effect instance, with no extra work. I know for sure catbird (https://github.com/travisbrown/catbird) has an Effect instance for Twitter Futures, and I believe monix and fs2 either have Effect instances or offer seamless conversions (I could be wrong here, I've never used either library).   

I'd be happy to look into how to provide instances for scalaz as well, although I've never worked with scalaz, it would be a good opportunity to learn.

Regarding Twitter Futures, they're certainly still used by anyone using Finagle/Finatra/Finch, and I think plenty of people are using those frameworks. As I mentioned, we may not need to offer a package if we point users to catbird -- users could simply bring in the elastic4s-cats-effect package and then use catbird to get to a Future. However, we could still offer an elastic4s-twitter package for users who don't want the cats / catbird dependencies.

sksamuel:
It all looks great. So my only real request now is to maybe rename FromListener into something that's a bit clearer that it's responsible for building the effect for execution.

daviswahl:
How do you feel about something like AsyncExecutor? And then the method would just be `async`. That was my original name, but I was concerned that both `async` and Execute are a bit overloaded already, with HttpExecutable and HttpRequestClient's async. 

Aside from that, I'll work on updating the relevant documentation and finishing out the tests. Once again, thank you for taking the time to consider this and being willing to take on the additional maintenance burden! I'll do whatever I can to help out with maintaining this code and any additional effect packages we end up adding.

sksamuel:
re: naming.

That's why I brought up the other PR. If you look at the potential refactoring there, with the `Request` class then `HttpExecutable` goes away, and perhaps this new typeclass of yours could fit in somehow. 

At the core level we need a request object that models the endpoint, params, url and body. Something that takes the request and sends it to elastic using the client. Something that takes the response and formats it into an object graph, and finally the FromListenerToBeRenamed which wraps the execution in whatever effect you need.

Could it somehow just be called an `Executable` (The TCP client will be deleted in 7.0 so don't worry about name clashes there). I'd have to think about it bit more, I don't have the code checked out.

The other I forgot to mention is that it has to work with `HttpRequestClient`, not `ElasticsearchJavaRestClient`, as people have written their own implementations if they don't want to use the apache http client provided by the elasticsearch java client.

daviswahl:
I looked over the commit and I see what you're getting at. Should we wait on this until that work is complete then? I'm also happy help out over there but I don't want to step on any toes.

sksamuel:
Mine was a proper WIP that I've barely begun, whereas this actually works. It will need to be changed to use HttpRequestClient or you're still leaking Futures out.

I could take your commit and run with it if you wanted, once you're happy with it.

daviswahl:
Sounds good, I'll let you know once things are cleaned up!

daviswahl:
Regarding using HttpRequestClient: I think I'm already doing that, unless I'm confused/mistaken. Can you sanity check when you have a chance?

sksamuel:
Your changes are specific to ElasticsearchJavaRestClient I think.

Ie, the listener interface is specific to the apache http client. If you were using say finagle, then you wouldn't have that interface.

 https://github.com/sksamuel/elastic4s/pull/1206/commits/4674538e928acce9d902f433b0bf16e7eb24f4bd#diff-ebe4f9b1879105ce70f091687ece3e37

daviswahl:
Oh, I see what you mean. So by parameterizing HttpRequestClient's async methods on FromListener we're actually breaking the interface and making it impossible to use with other http clients.

sksamuel:
Worst case it would break it, best case it just means your new effect stuff has no effect (pun not intended) on any other client.

daviswahl:
Ok, I think I get it. I'll add a dummy client for testing or get something going based on finatra. Not sure when I'll have time exactly but it shouldn't be more than a week or so.

daviswahl:
@sksamuel I think I'm happy with this. Check comments in the commits.

elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/ElasticsearchJavaRestClient.scala
```diff
@@ -14,20 +17,51 @@ class ElasticsearchJavaRestClient(client: RestClient) extends HttpRequestClient
...
+      HttpEntity.StringEntity(body, Some(contentEncoding))
+    }
+    val headers = r.getHeaders.map { header =>
+      header.getName -> header.getValue
+    }.toMap
+    HttpResponse(r.getStatusLine.getStatusCode, entity, headers)
+  }
+
+  private def doAsync[F[_]: AsyncExecutor](f: ResponseListener => Unit): F[HttpResponse] = {
+    AsyncExecutor[F].async { cb =>
```
sksamuel: Is this just a pattern?

Could you do,

```scala
private def doAsync[F[_]](f: ResponseListener => Unit)(implicit executor: AsyncExecutor[F[_]]): F[HttpResponse] = {
  executor.async { cb =>
````

Comment (daviswahl) [reply elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/ElasticsearchJavaRestClient.scala]: Yep that would work fine -- I have no preference. Would you like me to do that uniformly throughout the PR?

File: elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/AsyncExecutor.scala [hunk 2]
```diff
@@ -0,0 +1,47 @@
...
+import cats.Functor
+
+import scala.concurrent.{ExecutionContext, Future, Promise}
+
+/**
+  *  A typeclass that wraps an asynchronous computation.
+  *
+  *  Ideally, we'd paramterize on Functor instead of extending, but extending allows us to introduce this
+  *  without breaking API changes. TODO: In the next major release, F[_] should be parameterized on Functor.
+  */
```
sksamuel: Can you explain what you mean by F should be parameterized on Functor? F <: Functor or something else ?

daviswahl: I mean `[F[_]: Functor]` (or `implicit functor: Functor[F]`). Basically, cats (and cats-effect) provide typeclass instances of Functor for IO and Scala Future, so it's sort of redundant for us to provide our own instance. However, using them requires importing the appropriate typeclass implicits at the callsite, meaning users of elastic4s would need to update their code to add the imports. 

Requiring that AsyncExecutor implement its own Functor definition allows us to avoid this. However, it means we're doing extra work and have slightly more code to maintain. 

Honestly though, what I would do is remove the dependency on cats in core: It really ties us into an ecosystem that we're not leveraging beyond `Show`, which is something we can implement just fine without cats. Plus it makes providing a scalaz effect a little weird: scalaz users would be required to have a dependency on both cats and scalaz.

If we opted to pull cats out of core, I would define a `Functor` typeclass within elastic4s-http and then provide instances for it from cats/scalaz `Functor` in the cats/scalaz packages. Then we'd be fully agnostic to either ecosystem. 

Just my two cents.

sksamuel: I like your thinking, and I'm going to do just that.
Your changes are excellent, and they will get merged into master (with some changes to incorporate the other open PR). It'll be a week or two though as I've got other things to tend to at the moment, plus this kind of big change I want to get right and will be in 7.0.0 only not a 6.x release - but we can push out a 7.0.0-pre-pre-alpha release which is just 6.x.x in all but name.

daviswahl: Sounds great. Let me know if I can do anything to help.

Once this lands in master I'll make a PR with the cats-effect package I have sitting in another branch, and potentially look into the scalaz integration.

sksamuel: I did a cats effect, monix and scalaz implementation in my pulsar client.
https://github.com/sksamuel/pulsar4s
See what you think

It was easier to do over there as I had a few hours, this elastic change is going to take a lot longer as I want to change every request implicit.

sksamuel:
@daviswahl Hi Davis.

I've taken your work and used it as the basis for changes I've made in master. The work I've committed is more far reaching, as there were other changes that made sense to do together. The upshot is we (should) have the effect (pun not intended) you wanted, the ability to switch between Futures and other Scala effect types.

In addition, some small other improvements are (again following your excellent advice) to strip out cats from core/http modules, and having F parameterized on Functor.

If you have the time, please check out master, and see if what I've done still fits in with your requirements. I'm also happy to cite your authorship on the commits since it's the basis for the work.

daviswahl:
@sksamuel thanks for the update! I'm crazy busy over here but I'll get back to you in the next week.

sksamuel:
@daviswahl no rush, you've done the hard work already, so whenever you have a chance to take a look, no pressure :) Closing this in favour of the new ticket.

Issue #1195, opened 2017-12-22
Aws4ElasticClient does not refresh credentials


jnatten:
I use `Aws4ElasticClient` to query an aws elasticsearch endpoint.
This is using Elastic4s version `6.0.1` and Elasticsearch `6.0` in aws.
I create the client like this:
```
val elasticSearchUri = "elasticsearch://endpoint.com:443?ssl=true"
val client = Aws4ElasticClient(elasticSearchUri)
```

Which works fine if you are using permanent credentials, however if you are using temporary credentials (like the ones EC2 instances get), and the client sits for a while the aws credentials expire.
This makes calls to ElasticSearch fail with a 403 and message:
```
{
  "message": "The security token included in the request is expired"
}
```

I can't see a way to refresh the credentials of the client either so as of now i have to recreate the client every time the credentials expire.

saraiva132:
@jnatten I will look into that, it seems I might have to use a different provider that handles auto-refreshing or do it myself under the hood. Nice catch. Fixed your other issue btw, just bump the version to 6.1.1

MatteCarra:
I have the same problem. My code is running on an ec2 instance.

saraiva132:
@MatteCarra I was writing a solution with aws STS provider. I had/have the code ready but I was dealing with some issues with unitary tests because apparently STS Provider by default attempts to connect to a server (get initial session token I reckon), so without having a proper integration test I was not sure if it would fix the issue. 

@jnatten  Did #1207 fix this issue?

jnatten:
@saraiva132 It looks like the problem has been solved. 

I set up an instance with master (compiled at https://github.com/sksamuel/elastic4s/tree/db8f981f3269da48ceea251f7cb319d949f7c926) in EC2 and let it sit a while and the problem hasn't occurred.

PR #1207, opened 2017-12-30
Fix issue #1195

MatteCarra:
Method getCredentials is called every time a request is signed: this way new credentials are loaded when needed.


sksamuel:
:+1:

PR #1215, opened 2018-01-03
Fix #1212 metadata ignored by setter method of aggregation definitions

### LINKED ISSUE #1212: metadata ignored by setter method of aggregation definitions
trivial issue in all implementations of `metadata` setter method of *AggregationDefinition
`override def metadata(map: Map[String, AnyRef]): T = copy(metadata = metadata)`

artyomsol:


sksamuel:
:+1:

PR #1220, opened 2018-01-03
IpRange aggregation support in http client + test

lmafonso:
Another agg, when you have the time to review ... tks !


sksamuel:
Thanks again!

PR #1225, opened 2018-01-05
Search Type

bursalihilal:
Search types are added


sksamuel:
:+1:

bursalihilal:
Could you add this to tags/5.5.2 . I need to use 5.5.2 version.

sksamuel:
I can release another 5.5.x release, it won't be 5.5.2 as you can't change versions once released.

bursalihilal:
Which version will you do that and when could you do that? We have in a hurry a little.

bursalihilal:
I could not this version on Maven. Could you make it ?

sksamuel:
5.5.7 is on maven central now.

bursalihilal:
Thank you.

PR #1234, opened 2018-01-10
remove requires check for / for date math indexing

### LINKED ISSUE #1230: Date math in index names with createIndex
We're currently unable to use [date math in index names](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-rollover-index.html#_using_date_math_with_the_rollover_api) due to `CreateIndexDefinition` `require`ing that a `/` is not present in the name.
https://github.com/sksamuel/elastic4s/blob/master/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/indexes/CreateIndexDefinition.scala#L19

I was able to build a local version that worked correctly with the date math in the url by just removing the requires. 

I'm not sure what kind of approach you'd want to take.
We could expand the require to check if the `/` is in between `<` and `>` or just remove the check or implement date math checking?

I'd can make a PR implementing whatever approach sounds best

kalvinnchau:
This fixes #1230, since types are going to be going away in the next ES release.


sksamuel:
:+1:

PR #1239, opened 2018-01-12
Fix-1238 Add wait_for_no_relocating_shards param to cluster health and add timeout to request

tcdevoe:
Initially was just adding the `wait_for_no_relocating_shards` param and noticed `timeout` was not being included in the request params for cluster health requests, so I fixed that as well.


sksamuel:
Lovely thanks.

PR #1240, opened 2018-01-12
Fixed a bug with sort script parameters

Tomassino:
The problem was that all parameters were converted to string, losing their type. This, in turn, required a cast for parameters to the correct type inside scripts. I also added a simple test.


sksamuel:
Great thanks

PR #1243, opened 2018-01-14
Prevents NullPointerException when suggestion is not found

MatteCarra:
This change fixes NullPointerException when suggestion is not found and instead returns an empty list.


sksamuel:
:+1:

PR #1246, opened 2018-01-15
Use `index_patterns` instead of `template` in index templates

jcazevedo:
This PR modifies the way index templates are created to use `index_patterns` instead of `template`, since `template` is now a [deprecated field](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_60_indices_changes.html#_index_templates_use_literal_index_patterns_literal_instead_of_literal_template_literal).


sksamuel:
Thank you :+1:

Issue #1112, opened 2017-10-05
PutMappingDefinition supports rawSource but MappingDefinition does not


bryanmau1:
The `rawSource` field exists in `MappingDefinition` but is not used anywhere.  It was added for `PutMappingDefinition` and works there.  It almost seems like it was added to `MappingDefinition` on accident in this commit, but I might be missing code somewhere.

Could we have the `MappingBuilderFn` use `rawSource` just like `PutMappingBuilderFn` does?

https://github.com/sksamuel/elastic4s/commit/1bebf8606c68b09764ef85f9ac07eec3c96af30c

The use case is I'm trying to create a `CreateIndexDefinition` using `createIndex(<name>).copy(settings = settings, mappings = mappings)` and I create mappings as a `Seq[MappingDefinition]`from JSON

Issue #1071, opened 2017-09-07
MappingBuilderFn not using MappingDefinition.rawSource


jacobg:
MappingBuilderFn in the current master branch, and MappingContentBuilder in version 5.4.11, are both neglecting to check MappingDefinition.rawSource.

PR #1247, opened 2018-01-16
Use rawSource to build mapping if provided

mrszg:
This PR adds support for rawSource in MappingDefinitionLike. It is related to issue   #1112 and #1071


mrszg:
I think that travis tests failures are not deterministic and not related to this PR (repeated tests on forked repo succeded)

sksamuel:
:+1:

PR #1248, opened 2018-01-17
Support extracting geobounds and geocentroid aggregation results

mrszg:
This PR adds support for extracting geo_bounds and geo_centroid aggregation results (from http responses).


sksamuel:
:+1:

PR #1252, opened 2018-01-18
Upgrade SBT to 1.1.0

joan38:


build.sbt
```diff
@@ -194,7 +200,7 @@ lazy val docsMappingsAPIDir = settingKey[String]("Name of subdirectory in site t
 lazy val docs = project
   .in(file("docs"))
   .enablePlugins(MicrositesPlugin, ScalaUnidocPlugin)
-  .settings(ghpages.settings)
+//  .settings(ghpages.settings)
```
joan38: This is gone, what should I do?

project/Build.scala
```diff
@@ -35,22 +35,21 @@ object Build extends AutoPlugin {
   override def projectSettings = Seq(
     organization := org,
     // a 'compileonly' configuation
-    ivyConfigurations += config("compileonly").hide,
+//    ivyConfigurations += config("compileonly").hide,
```
joan38: `config must be directly assigned to a val, such as `val Tooling = config("tooling")`.`
What is this setting doing?

project/plugins.sbt
```diff
@@ -1,9 +1,9 @@
 resolvers += Classpaths.sbtPluginReleases
 
+addSbtPlugin("org.scalariform" % "sbt-scalariform" % "1.8.2")
```
joan38: Can we move onto a real code formatter akka Scalafmt? I'm happy to do a PR after this one if we agree one this.

sksamuel:
As long as the rules are set to how I like to write code. So either remove
it completely or setup scalafmt to match my style:)

build.sbt
sksamuel: Remove it

project/Build.scala
sksamuel: No idea !

joan38: Is it ok to remove this then?

sksamuel:
I think so

joan38:
@sksamuel This PR should be good to go then

sksamuel:
I've had to revert this in master, because when I try to release it is adding "scm" tags twice (that's what maven central is complaining about). Until I figure it out we're back on sbt 0.13.

joan38:
Where do you get this double "scm"?
Do you have a way to reproduce for me without all your maven credentials?

sksamuel:
No I don't think so, the error is within oss.sonatype.org when I try to do the release.

joan38:
Also I've seen you tried to add back:
```
    // a 'compileonly' configuation
    ivyConfigurations += config("compileonly").hide,
    // appending everything from 'compileonly' to unmanagedClasspath
    unmanagedClasspath in Compile ++= update.value.select(configurationFilter("compileonly")),
```
Do you know what is this used for?

sksamuel:
Nope, but I thought it might fix it. It doesn't.
It might be the case that sbt 1.1 just adds SCM to the pom automatically in which case it could be removed from the declared pom in the build file.

joan38:
It seems that I can't help much on this, so I let you experiment.
Let me know if I can be of any help.
Also https://github.com/sksamuel/elastic4s/pull/1254 will be blocked until we move on SBT >=1.0 as there is no support for SBT 0.13.

sksamuel:
I suppose I can un-revert back to sbt 1.1.0 now, as I won't need  to do a release for a week or so.

joan38:
May you can try a revert of the revert, deleting the scm from the pom and try if this works locally?
And If that works commit the change.

joan38:
See: https://github.com/sksamuel/elastic4s/pull/1270

PR #1257, opened 2018-01-22
Separate HTTP request marshallers from executables

catalin-ursachi:
Addresses comment https://github.com/sksamuel/elastic4s/pull/1185#discussion_r158583449 in #1185

### DIFF
--- a/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/HttpExecutable.scala
+++ b/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/HttpExecutable.scala
@@ -1,6 +1,7 @@
 package com.sksamuel.elastic4s.http
 
 import com.sksamuel.exts.Logging
+import org.apache.http.entity.ContentType
 
 import scala.concurrent.Future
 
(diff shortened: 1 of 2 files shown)


catalin-ursachi:
I've called them "Marshallers" rather than "Builders" so I could call the result `MarshalledRequest` (but I suppose `BuiltRequest` or simply `Request` would also work, if you prefer either of those).

sksamuel:
I've taken this and incorporated it into master as part of a larger change. Now `show` will include endpoint, method and parameters.

catalin-ursachi:
Ah, great. I suppose this and #1185 can be closed, then?

sksamuel:
Assuming I've not missed anything out you'd like to see.

catalin-ursachi:
No, all looks good to me! 🙂 

Only suggestion I'd have is renaming the method:

```handler.requestHander(t: T): ElasticRequest```

since it doesn't return a handler, but a marshalled request. Maybe `handleRequest`, `buildRequest`, or `marshalRequest` would be clearer?

sksamuel:
Yes good point.

PR #1258, opened 2018-01-24
Add arrays to autovalue

joan38:


sksamuel:
Thanks

joan38:
BTW is there a reason for accepting such a generic type in for exemple `termsQuery`?
Is there a reason for not getting `Iterable[String]`?

PR #1266, opened 2018-01-25
Bucket script

F-BB:
Currently we can not add bucket script in pipeline type aggregations
My PR adds this opportunity


sksamuel:
:+1:

Issue #1244, opened 2018-01-14
Checking whether avg aggregation has null value


Tomassino:
When executing avg aggregations the resulting "value" may be null (e.g. because no document matches the query). In this case accessing the result of the aggregation using `result.aggregations.avg("name")` caused a NullPointerException to be thrown. The problem is at [this point](https://github.com/sksamuel/elastic4s/blob/2caafd489b7eceaceabee07467fddb8423a3564e/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/search/aggresponses.scala#L277), that is:

```scala
def avg(name: String): AvgAggResult = AvgAggResult(name, agg(name)("value").toString.toDouble)
```

which calls `.toString` on "value" without any check (some other metric aggregations use the same code, which I suppose leads to the same problem, but I've not tested them). Other metric aggregations (e.g. min or max) use an Option as the value, which solves the problem. As a workaround, in my code I had to do something like this:

```scala
val aggs = r.aggregations
if (aggs.data("avg_duration").asInstanceOf[Map[String, Any]]("value") != null)
  aggs.avg("avg_duration").value
else
  0.0
```

Am I missing something? In my opinion the best solution to this issue would be to make the `value` field in `AvgAggResult` an `Option[Double]`, but this would mean changing the API. Alternatively one could add a method to the `HasAggregations` trait that checks whether "value" is null or not (`def valueIsNull: Boolean`?).

If you decide this is something to fix I can work on a patch, in this case let me know how to fix it

sksamuel:
Maybe we could change the result case class to be optional, but add in a method with the old name that throws an exception ?

Tomassino:
Sorry for the late reply. Are you suggesting to do something like this?

```scala
case class AvgAggResult(name: String, valueOpt: Option[Double]) extends MetricAggregation {
  def value: Double = valueOpt.get
}
```

Or perhaps you mean changing the `HasAggregations` trait so that it has both `def avg(name: String): AvgAggResult` and `def avgOpt(name: String): Option[AvgAggResult]`?

I personally prefer the first option (change `AvgAggResult`) because it also makes that class more like the other metric aggregations results (e.g. min and max)

PR #1271, opened 2018-01-29
Change AvgAggResult, SumAggResult and ValueCountResult to have optional values

Tomassino:
The new optional value is called `valueOpt` while `value` simply performs a `.get` on the option, possibly causing an exception to be raised

This is a proposed solution to issue #1244


sksamuel:
Very nice :+1:

PR #1273, opened 2018-01-29
Fixed NoSuchElementException when getting mapping without properties

mrszg:
Sometimes mapping doesn't have "properties". This PR fixes NoSuchElementException thrown when getting mapping without properties.


sksamuel:
:+1:

PR #1282, opened 2018-02-06
use DefaultAwsRegionProviderChain to select default aws region

dmjb:
Right now, the `elastic4s-aws` code checks the `AWS_DEFAULT_REGION` environment variable to find the default AWS region. This is only one of the techniques which the AWS SDK uses to determine the default region, and thus this code fails in scenarios where the AWS SDK can succeed (e.g. I have some code which uses the EC2 metadata service to find the region)

Use the `DefaultAwsRegionProviderChain` class from the AWS SDK to find the default region. This encapsulates the various default region-finding rules as described in the documentation: https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/java-dg-region-selection.html


sksamuel:
:+1:

Issue #1278, opened 2018-01-31
Why switched return types of HttpClient.execute to Either?


polarke:
Response[U] seems more in Scala way than Either[RequestFailure, RequestSuccess[U]]

sksamuel:
`scala.Either` in the scala SDK seems pretty Scala to me :)

aloiscochard:
Hey @sksamuel, I think that was @polarke was trying to say is that encoding anything as `Either[RequestFailure, RequestSuccess[U]]` is simply nonsensical.

It's kinda like encoding a boolean as `Either[Unit, Unit]`.

Look in `HttpClient.scala` you define an awesome algebraic data type `sealed trait Response[+U]`.

It is sealed! That is awesome! That give us exhaustive checking :-)

It means that if I have a `Response` I can simply pattern match on it and treat the two different cases:
- `RequestFailure`
- `RequestSuccess`

In that regards, wrapping such response in an `Either` is just adding structural noises!
Or do you think this add any value? If yes, please show an example.

If you agree with the thinking described here, that it is non nonsensical to wrap a `Response` in an `Either`, we will be happy to provide a PR.

sksamuel:
@aloiscochard what you have said is of course completely sensible. I have tried to remember why I did it with the ADT _and_ the Either and I am afraid I cannot remember. Therefore I must assume I was either drunk or being an idiot.

There's no need to provide a PR because I need to change this kind of stuff anyway to incorporate #1257 and #1206 and #1885. All four issues (those three plus this one) I am going to fix in a single unified far reaching commit, which will be released as part of 6.2 or 7.0

aloiscochard:
@sksamuel thanks for the detailed answer! my gut feeling is that you wanted to use Either to get some facilities (syntax...) associated with it at some point without realizing the ADT should be enough.

I often discussed with friends the possibility of a USB device that would record my alcoholic blood level directly in my git commit messages... Ballmer peak driver programming!

Anyway, thanks again for taking feedback into account and supporting such a useful library.

sksamuel:
Closing this in favour of the new ticket #1279

sksamuel:
@aloiscochard @polarke  let me ask for your feedback if I may.

[...]

PR #1293, opened 2018-02-21
Fix `Response.map` and add `Response.flatMap`.

aloiscochard:
In response to your question on #1278, here is a PR that introduce the required method (with proper signature, which means breaking change for people using `map` previously... I introduced `toOption` that gives back similar feature) to provide the for-comprehension syntax.

I did include a test, that exercise the feature.

I hope you'll like it :-)


sksamuel:
Awesome ! :+1:

PR #1296, opened 2018-02-22
Respect IndexRequest.createOnly when used inside bulk through HTTP - for 5.5.x

lustefaniak:
Backport https://github.com/sksamuel/elastic4s/pull/1294


sksamuel:
:+1:

PR #1297, opened 2018-02-27
fixed edge_ngram name

touchdown:
looks like it should be "edge_ngram" according to this https://www.elastic.co/guide/en/elasticsearch/guide/current/_index_time_search_as_you_type.html

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -3,6 +3,7 @@ import microsites.ExtraMdFileConfig
 lazy val root = Project("elastic4s", file("."))
   .settings(
     publish := {},
+    publishLocal := {},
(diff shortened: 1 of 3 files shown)


sksamuel:
Thanks for the PR. Can you fix up the errors in CI before I merge?

touchdown:
ok so this apparently has `edgeNGram` https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-edgengram-tokenfilter.html
so I m going to do a bit more digging before submit my final commit

touchdown:
ok seems like current one is working fine, was a different issue originally, closing

PR #1298, opened 2018-02-27
Fix broken termsQuery when empty seq is passed

TrustNoOne:
Hi, termsQuery currently fails with an empty Seq. I have a dynamic termsQuery in my code, and sometimes it filters by "Seq()", that's how I found out.  This should fix it: it uses the ref lookup if specified otherwise fallbacks to normal terms instead of viceversa.

cheers


sksamuel:
:+1:

PR #1309, opened 2018-03-15
Fix a problem on LocalNode closing. Fix #1310

### LINKED ISSUE #1310: The jvm will not exit after the LocalNode close.
When I close the client like this:

```scala
val localNode = LocalNode("findex0303", "./esTmp/tmpDataPath0303")
val client = localNode.http(true)
client.close()
```

the jvm will not be close.
and this is the successful code to fix this problem.

```scala
val localNode = LocalNode("findex0303", "./esTmp/tmpDataPath0303")
val client = localNode.http(true)
client.close()
client.client.close()
```

djx314:
When the ElasticClient generated by LocalNode closed, the HttpClient will not be closed. That leads the jvm will not close automatically. This pr fix the problem this problem by close the HttpClient on ElasticClient closing. Fix #1310


sksamuel:
:+1:

PR #1311, opened 2018-03-17
allow AWS signer to be created from creds provider

dmjb:
I have a use case where I need `Aws4RequestSigner` to use a custom `AWSCredentialsProvider` (specifically, I have code which requires use of the Security Token Service) Currently the code accepts an access/secret key pair, and constructs a basic `AWSCredentialsProvider` behind the scenes based on those keys.

In order to address my requirements, and maintain backwards compatibility, I have taken the following approach:

1) Create a separate configuration case class which accepts an `AWSCredentialsProvider` instead of a access/secret key pair.
2) Create a method to allow the original config case class to be converted into the implementation I have added.
3) Modify the private classes to accept the implementation which I have added.

I'm not really sure about the `Aws4ElasticConfigWithProvider` name, better suggestions welcome.


dmjb:
@sksamuel any blockers to getting this merged?

rossmckelvie:
@sksamuel AWS instances have a credentials provider on them, which allows you to use instance profiles that are attached to the instance to connect to AWS's Elasticsearch Service. Using the default credential provider chain is the [recommended method in AWS docs](https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html)

Behind the scenes, the instance credentials provider is responsible for providing & refreshing temporary tokens that use the Security Token Service described in @dmjb's description. It would be great if example usage for AWS was included in the README, or a README within the elastic4s-aws module. If this gets merged in and we are able to use this library I'm happy to help start those docs.

PR #1320, opened 2018-03-28
Fix a bug in build result to case class by circe.

djx314:
When use `(response: RequestSuccess[GetResponse]).result.safeTo[CommonBody]` to decode the result to case class use circe.
If the `_id` is not found. The `GetResponse._source` will be null. Then the `result.safeTo` will throw a uncatched NullPointException. Not the excepted Left[Exception].
This pr hope for fix it.


sksamuel:
Great thanks.

sksamuel:
If you need this adding to a branch release, do a PR to the release branch and I'll do a release onto maven.

PR #1321, opened 2018-03-28
Fix AWS sign with unordered query param

lucaagostini:
If the URL to sign hasn't the query params already ordered the AWS sign is not correct.
If I specify indices options the resulting URL hasn't the query param ordered.
The fix is to build the canonicalQueryString with ordered query params.

https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html 
To construct the canonical query string, complete the following steps:
a) Sort the parameter names by character code point in ascending order. For example, a parameter name that begins with the uppercase letter F precedes a parameter name that begins with a lowercase letter b.

Related to #1326


sksamuel:
:+1:

sksamuel:
Can you add the same to master as well please ?

lucaagostini:
Yes, sure.
I've just created #1328

PR #1322, opened 2018-03-28
Add support for custom AWS auth chain

lucaagostini:
At the moment it's not possible to sign requests to ElasticSearch with a custom auth chain.
The only allowed ways are:
* with static access_key/secret_key
* with the default AWS auth chain

In some cases it could be usefull set the auth chain: for example if I need to assume a role to access ElasticSearch (`new STSAssumeRoleSessionCredentialsProvider.Builder(myrole,mysessionanme).build` instead of using the user permission). 
This scenario is common when someone need to access a cluster that is in a different AWS account of the user/role that wants to access it.

### DIFF
--- a/elastic4s-aws/src/main/scala/com/sksamuel/elastic4s/aws/Aws4ElasticClient.scala
+++ b/elastic4s-aws/src/main/scala/com/sksamuel/elastic4s/aws/Aws4ElasticClient.scala
@@ -1,13 +1,13 @@
 package com.sksamuel.elastic4s.aws
 
+import com.amazonaws.auth.{AWSCredentialsProvider, AWSStaticCredentialsProvider, BasicAWSCredentials, DefaultAWSCredentialsProviderChain}
+import com.amazonaws.regions.DefaultAwsRegionProviderChain
 import com.sksamuel.elastic4s.ElasticsearchClientUri
 import com.sksamuel.elastic4s.http.HttpClient
-import com.amazonaws.auth.{AWSStaticCredentialsProvider, BasicAWSCredentials, DefaultAWSCredentialsProviderChain}
-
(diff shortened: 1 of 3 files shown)


sksamuel:
Does this clash with #1311 
I don't know anything about AWS unfortunately.

lucaagostini:
The #1311 seems to do the same thing.
What I could suggest is to avoid to set the `service` attribute in the config class (even from the already existing Aws4ElasticConfig) because it cannot assume correctly any other values than `es` (it's the AWS service code name you are signing and with elastic4s we can sign only requests to ElasticSearch)

lucaagostini:
To be able to access the `requestConfigCallback` parameter in the client I've changed the `Aws4ElasticClient` class and put as public the `AwsHttpClientConfigCallback` class that extends `HttpClientConfigCallback` for authentication.

Probably the best way to create the `HttpClient` with AWS auth support (with this solution) is to pass the httpCallback directly in the classic `HttpClient` instead of using the `Aws4ElasticClient` class. In this way, if the `HttpClient` will change the signature the AWS authentication doesn't require changes.

For example:

```
HttpClient(
  uri,
  requestConfigCallback = requestConfigCallback,
  httpClientConfigCallback = new AwsHttpClientConfigCallback(credentialProvider, region)
)
```

@sksamuel if you prefer I can remove the `Aws4ElasticClient` class

sksamuel:
I will try and review this soon. There are two outstanding PRs for AWS which I haven't merged as I don't know much about AWS and I need to quickly brush up before I merge.

hitotsukami:
@sksamuel I have reviewed and it looks good to me. +1

masonkirchner:
@sksamuel When can we get this merged?

Issue #1323, opened 2018-03-28
Support method (algorithm) when defining percentiles aggregation (v5.5.x)


pgorin:
PercentilesAggregationDefinition is not currently supporting underlying algorithm or method definition on version 5.5.x of elastic-core module.

I will make a PR to add support for TDigest and HDR methods.

pgorin:
Can you make a release of the 5.5.10 version? Thanks a lot!

PR #1324, opened 2018-03-28
Added support for underlying method when building Percentiles aggregation definition.

pgorin:
Related to issue #1323


pgorin:
Sorry, I have found that using percentileRanksAggregation solves the problem.

sksamuel:
Would be worth putting on master too if it's not already.

PR #1325, opened 2018-03-28
Use GET rather than POST for multiget

Tyruiop:
To use the verb suggested here: https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-multi-get.html


sksamuel:
If you need this soon, do the same PR to a release branch of the version you want and I'll do a release.

Tyruiop:
Thanks, I might do that then, when do you think the next release will be otherwise?

sksamuel:
Master branch will be used for 7.0.0 as it has some breaking changes which
is whenever elastic release it. Could be weeks months.

PR #1328, opened 2018-03-29
Fix missing query param sort in canonicalQueryString in the AWS sign

lucaagostini:
The same of #1321 but for the master branch


sksamuel:
Thanks

tkaiser:
still seeing this in 6.3.7 when doing a _cat/indices call

> URI [/_cat/indices?v&format=json&bytes=b], status line [HTTP/1.1 403 Forbidden]
{"message":"The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details."}

PR #1335, opened 2018-04-02
#1320 to 6.2.x branch

djx314:
Hi, here is the pull request. I only create the pr for 6.2.x because I only use v6.2.3. I am afraid that I don't know the different between 6.2.x and other branches.


djx314:
It seems that some other files is broken?

PR #1341, opened 2018-04-04
Fix broken termsQuery when empty seq is passed

grillorafael:
Is a PR like this that you suggested?

#1338

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ scala:
 - 2.12.4
 
 script:
-- sbt clean scalafmtCheck scalafmtSbtCheck test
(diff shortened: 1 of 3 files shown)


sksamuel:
Maven central is being slow atm, I'll keep trying to release but it might be tomorrow.

grillorafael:
I just updated the PR disabling travis at the same time you merged. Please make sure this is a valid change!

sksamuel:
Yep.

grillorafael:
Thanks! Tomorrow would be just fine if you can't today. And thanks for the fast feedback as usual

PR #1342, opened 2018-04-05
Fixed GeoHashGridAggregation key value selection.

Kamill0:
Following PR changes the way the key value is being selected from map while constructing the GeoHashGridBucket object.

Before the change:
`GeoHashGridBucket(Some("u173z"),...`

After:
`GeoHashGridBucket("u173z",...`


mrszg:
+1

sksamuel:
:+1:

PR #1343, opened 2018-04-07
Fix join mapping type

chriso:
This PR includes a fix for https://github.com/sksamuel/elastic4s/pull/1191, and adds support for more than one child per parent.

According to [the docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/parent-join.html), the mapping type should be `join` rather than `object`, and although a mapping may have only one type, the docs say:

> An element can have multiple children but only one parent.

For example:

```
~ $ curl -X PUT http://localhost:9200/my_index -H 'Content-Type: application/json' \
    -d '{"mappings":{"_doc":{"properties":{"type":{"type":"join","relations":{"parent":["child1","child2"]}}}}}}'
{"acknowledged":true,"shards_acknowledged":true,"index":"my_index"}
```

```
~ $ curl -s http://localhost:9200/my_index | jq                                                                                                                                                 
{
  "my_index": {
    "aliases": {},
    "mappings": {
      "_doc": {
        "properties": {
          "type": {
            "type": "join",
            "eager_global_ordinals": true,
            "relations": {
              "parent": [
                "child2",
                "child1"
              ]
            }
          }
        }
      }
    },
    "settings": {
      "index": {
        "creation_date": "1523059422053",
        "number_of_shards": "5",
        "number_of_replicas": "1",
        "uuid": "WgZ4-v2uRk687w6RY4lGTQ",
        "version": {
          "created": "6020399"
        },
        "provided_name": "my_index"
      }
    }
  }
}
```

### DIFF
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/mappings/CommonFieldBuilder.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/mappings/CommonFieldBuilder.scala
@@ -99,7 +99,7 @@ object FieldBuilderFn {
         builder.startObject("relations")
         join.relations.foreach {
           case (parent, child) =>
-            builder.field(parent, child)
+            builder.autofield(parent, child)
         }
         builder.endObject()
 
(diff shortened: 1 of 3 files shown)


sksamuel:
:+1:

chriso:
Should I port this fix to the `release/6.2.x` branch? I just read on another PR that master won't be released until Elasticsearch 7 is released.

chriso:
Actually nm, I've just realised that we can use `MappingDefinition.rawSource` to bypass the DSL limitation.

PR #1345, opened 2018-04-09
Added filter_duplicate_text to significant text aggregation

tcdevoe:


tcdevoe:
Not sure why `NestedQueryTest`  failed - it shouldn't be affected by the changes I made and it only failed for scala 2.12, jdk 8. Seems like maybe it's an unreliable integration test?2

sksamuel:
I've kicked the test off again.

sksamuel:
Passed this time.

Issue #1156, opened 2017-12-07
TopHits don't have sourceAsString or any easy method to parse to a Type


grillorafael:
Hey @sksamuel 

I'm migrating here some of the code to use `6.0.0` and I'm also changing from `elastic4s-tcp` to `elastic4s-http` as advised.

I saw that https://github.com/sksamuel/elastic4s/blob/master/elastic4s-tcp/src/main/scala/com/sksamuel/elastic4s/searches/aggs/RichAggregations.scala does not exist in `elastic4s-http` but only in the `tcp` one.

Is there any specific reason for that?

The code I was using before was the following:

```scala
val filterAggregation = response.aggregations.getAs[Filter](filterAggName)
```
and It seems that I can't import things like `org.elasticsearch.search.aggregations.metrics.avg.InternalAvg` when using the HTTP library

I have the following modules:

```scala
val esversion = "6.0.0"

"com.sksamuel.elastic4s" %% "elastic4s-core"  % esVersion
"com.sksamuel.elastic4s" %% "elastic4s-http" % esVersion
"com.sksamuel.elastic4s" %% "elastic4s-http-streams"  % esVersion
"com.sksamuel.elastic4s" %% "elastic4s-play-json"  % esVersion
```

grillorafael:
I think I managed to understand how to do it now.

```scala
val filterAggregation = response.result.aggs.filter(filterAggName)
val groupByAggregation = filterAggregation.terms(byResolvedIMOAggName)
```

Seems ok the issue now is that I have a `topHits` aggregation and before I was doing `Json.parse(h.getSourceAsString).as[LiveVessel]` on the `Hit`. Now I don't have access to this `getSourceAsString` and it seems that I only have access to `source` as a `Map[String, Any]`

any better way to do this?

grillorafael:
`Json.toJson(h.source).as[LiveVessel]` seems ok

grillorafael:
I'm going to close this issue as it was just me being a bit dumb in the morning.

grillorafael:
`Map[String, Any]` is not a great type for PlayJson to parse. Might be better to have access to the raw string or something similar. Like `to[]` or `safeTo[]` that we have on the other ones

sksamuel:
In 6.0.0 + the response types include the raw json as well, so you can always get the json yourself if you want to. I'm going to add `to` for aggregations as well.

sksamuel:
Added to and safeto for aggs in 6.0.4.

[...]

PR #1358, opened 2018-04-23
feat: 'readable' TopHit and TermAgg

nolledge:
As I wanted to parse a single TopHit of an aggregation to a custom type I only found the contained data as Map[String, Any] on the case class.
 
This is not a very convenient type to parse as described in [this issue](https://github.com/sksamuel/elastic4s/issues/1156).

On the aggegation containers there was already a method to transform data. I moved it to a trait and made it available in TopHit and TermAgg


sksamuel:
:+1: 

If you want this releasing sooner than 7.0.0 please also PR to release/6.2.x

Issue #1359, opened 2018-04-23
Exception when deserializing response to nested query with disabled source


pdorobisz:
Executing query with disabled source for `innerHits`:

```
myQuery.inner(innerHits("myInnerHits").fetchSource(FetchSourceContext.DO_NOT_FETCH_SOURCE))
```

will throw following exception:

```
key not found: _source
java.util.NoSuchElementException: key not found: _source
	at scala.collection.MapLike$class.default(MapLike.scala:228)
	at scala.collection.AbstractMap.default(Map.scala:59)
	at scala.collection.MapLike$class.apply(MapLike.scala:141)
	at scala.collection.AbstractMap.apply(Map.scala:59)
	at com.sksamuel.elastic4s.http.search.SearchHit$$anonfun$innerHits$2$$anonfun$apply$1.apply(responses.scala:48)
	at com.sksamuel.elastic4s.http.search.SearchHit$$anonfun$innerHits$2$$anonfun$apply$1.apply(responses.scala:44)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
	at scala.collection.immutable.List.foreach(List.scala:392)
	at scala.collection.TraversableLike$class.map(TraversableLike.scala:234)
	at scala.collection.immutable.List.map(List.scala:296)
	at com.sksamuel.elastic4s.http.search.SearchHit$$anonfun$innerHits$2.apply(responses.scala:44)
	at com.sksamuel.elastic4s.http.search.SearchHit$$anonfun$innerHits$2.apply(responses.scala:39)
	at scala.collection.MapLike$MappedValues.get(MapLike.scala:249)
	at scala.collection.MapLike$class.apply(MapLike.scala:140)
	at scala.collection.AbstractMap.apply(Map.scala:59)
```

PR #1361, opened 2018-04-24
Fix exception when inner hit's source not included in response 

pdorobisz:
#1359


sksamuel:
:+1:

pdorobisz:
When will it be released as a new version (5.5.11)?

sksamuel:
5.5.11 is released.

PR #1327, opened 2018-03-29
Support script in Reindex api

TrustNoOne:
just that

cheers

PR #1362, opened 2018-04-25
Add script to reindexfn on 6.2.x

TrustNoOne:
ref https://github.com/sksamuel/elastic4s/pull/1327


TrustNoOne:
thanks! I really needed this because it broke the scripts we run on automatic reindexing when we upgrade index mappings

PR #1363, opened 2018-04-26
Upgrade Scalafmt

joan38:
This upgrade Scalafmt to 1.5 which:
- Add support for sorting modifiers. Making them consistant across the codebase
- Removes the useless brackets everywhere and not only in functions

Let me know if the change is not crazy.


sksamuel:
I quite like the useless brackets :)

joan38:
Thanks @sksamuel

PR #1370, opened 2018-05-07
Port [#1294] Respect IndexRequest.createOnly when used inside bulk through HTTP to 6.2.x branch

lustefaniak:
Looks like I missed `release/6.2.x` branch last time when doing https://github.com/sksamuel/elastic4s/pull/1294, now when trying to use 6.2.6 I got hit by that problem again.

Should I also create PR to 6.1.x ? Or is there any better way how to handle that?


lustefaniak:
Also not all elements of 6.2.7 got released, eg. `com.sksamuel.elastic4s:elastic4s-embedded_2.12:6.2.7` and `com.sksamuel.elastic4s:elastic4s-testkit_2.12:6.2.7` are missing.

sksamuel:
I don't know why 6.2.7 is missing some components. I've released 6.2.8 to address this + latest commits.

PR #1371, opened 2018-05-07
Make sure bulk deletes contain `version` field (master)

lustefaniak:


sksamuel:
Thanks for doing both 6.2.x and master.

Issue #1376, opened 2018-05-09
Support Composite Aggregations


markuspeters:
Composite aggregations are multi-bucket aggregations that create one bucket for potentially every combination in the cross-product of values from one or more fields. The functionality is in beta as of Elasticsearch 6.2, further details are in the [Elasticsearch docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-composite-aggregation.html).

As an elastic4s user, I would like to be able to express such aggregations using the usual DSL, e.g.,
```
SearchDefinition("myindex" / "mytype").aggs(
    CompositeAggregationDefinition("comp", 
      sources = Seq(
        TermsValueSource("s1", field = Some("f1"), order = Some("desc")),
        HistogramValueSource("s2", 5, field = Some("f2"), order = Some("desc")),
        DateHistogramValueSource("s3", "5d", field = Some("f3"), order = Some("desc"), timeZone = Some("utc"))
      )))
```

markuspeters:
I'm working on a pull request against this issue.

sksamuel:
Excellent, a PR would be great. Please can you PR against 6.2.x branch if you want a release doing, and also against master.

markuspeters:
Thank you @sksamuel. I submitted the PR for the 6.2.x just now. If you find the PR acceptable I'll submit a second one against master.

sksamuel:
6.3.0 is now released.

PR #1377, opened 2018-05-09
Add support for composite aggregations to 6.2.x #1376

markuspeters:
Composite aggregations are multi-bucket aggregations that create one
bucket for potentially every combination in the cross-product of values
from one or more fields. This commit introduces DSL elements to express
composite aggregations such as the following:

SearchDefinition("myindex" / "mytype").aggs(
  CompositeAggregationDefinition("comp", sources = Seq(
    TermsValueSource("s1", field = Some("f1"), order = Some("desc")),
    HistogramValueSource("s2", 5, field = Some("f2")),
    DateHistogramValueSource("s3", "5d", field = Some("f3"),
                             timeZone = Some("utc"))
  )))


sksamuel:
:+1:

PR #1378, opened 2018-05-11
fix: correctly add bounding box options

grillorafael:
Bounding box options were added within `locationField` node instead of within` geo_bounding_box`


sksamuel:
:+1:

Issue #1368, opened 2018-05-04
[HTTP DSL] Index exists check handles error responses inproperly


unkarjedy:
(in context of version 5.4)

There is "index exists" method in ```com.sksamuel.elastic4s.admin.IndexAdminApi```:
```scala 
def indexExists(index: String): IndexExistsDefinition = IndexExistsDefinition(index)
```
and corresponding HttpExecutable in  ```com.sksamuel.elastic4s.http.index.admin.IndexAdminImplicits```:
```scala
implicit object IndexExistsHttpExecutable extends HttpExecutable[IndexExistsDefinition, IndexExistsResponse] {
    override def execute(client: RestClient, request: IndexExistsDefinition): Future[IndexExistsResponse] = {
        val endpoint = request.index
        logger.debug(s"Connecting to $endpoint for indexes exists check")
        val resp = client.performRequest("HEAD", endpoint)
        Future.successful(IndexExistsResponse(resp.getStatusLine.getStatusCode == 200))
    }
}
```

The common usage would look like this:
```scala
http.execute(indexExists(name)).flatMap(_.isExists).map { exists =>
    if(exists) {
        Future.successful(Unit)
    } else {
        http.execute(createIndex(indexName).source(mapping)).map(_ => Unit)
    }
}
```

But unfortunately this code can lead to "index already exists" error response from elastic.
This is so because the test for ES response code in not proper:
```scala 
resp.getStatusLine.getStatusCode == 200
```
In our project ES was facing some temporary internal issues and returned bad request 400 when checking whether index exists. But the index extested! So the above test returned `false` and our app tried to recreate already existing index.

The only valid case of "index does not exists" should be treated when ES returns 404 code.
I would suggest replacing the above line with something like this:
```scala
resp.getStatusLine.getStatusCode match {
    case 200 => Future.successful(IndexExistsResponse(true))
    case 404 => Future.successful(IndexExistsResponse(false))
    case code => Future.failed(new ElasticsearchException(s"Bad response code: $code"))
}
```
but i am not sure which exception would be better in according to your style/standarts

unkarjedy:
@sksamuel If you agree I could preapare pulrequests.

sksamuel:
Yes that seems reasonable, a PR would be good.

unkarjedy:
@sksamuel I've prepared pullrequest for version 5.4
[https://github.com/sksamuel/elastic4s/pull/1379 ](url)
if I have enough time next week i will review other release versions

[...]

PR #1379, opened 2018-05-11
#1368 treat only 404 response code as "does not exists" for indexes/types/aliases

unkarjedy:
see [https://github.com/sksamuel/elastic4s/issues/1368](url)


sksamuel:
:+1:

PR #1380, opened 2018-05-16
style: typo delete api document

letusfly85:
I changed `delete.md`.
Because following way,`delete` method, does not work. And it will be compiled error.

```scala
  client.execute {
    delete id 3 from "places" / "cities"
  }
```


sksamuel:
:+1:

letusfly85:
@sksamuel 

Thank you!

PR #1381, opened 2018-05-18
Build for scala 2.12.6

sebastiaansamyn-tc:


sebastiaansamyn-tc:
We have a project that is still using elasticsearch 1.5 without the possibility to upgrade soon. With scala 2.12 being the latest stable version, we'd like to cross-build the 1.5.x dsl to scala 2.12 so we can upgrade this project to a more recent scala version.

sksamuel:
Had to upgrade the build to sbt 1.1.1 and fix the release procedure, it's been a while since this branch was released and sbt/scala/plugins have moved on a bit since then.

sebastiaansamyn-tc:
Many thanks!

PR #1395, opened 2018-05-26
Fix missing suggestions response

benniekrijger:
When using the suggester API and no results are returned, the "suggest" field is missing from the response. This results in a NullPointerException at the moment, ex:
```
java.lang.NullPointerException: null
at com.sksamuel.elastic4s.http.search.SearchResponse.suggestion(responses.scala:109) ~[com.sksamuel.elastic4s.elastic4s-http_2.12-6.2.9.jar:6.2.9]
at com.sksamuel.elastic4s.http.search.SearchResponse.completionSuggestion(responses.scala:117) ~[com.sksamuel.elastic4s.elastic4s-http_2.12-6.2.9.jar:6.2.9]
```


sksamuel:
:+1:

PR #1399, opened 2018-06-05
https://github.com/elastic/elasticsearch/issues/6299

dpblh:
was added pagination to top_hits aggregation.
it's very small changes. just added the possibility of pagination.
i'd rather to see the changes and in previous version.


sksamuel:
:+1: 

Would be great to have a PR for this to master as well.

Issue #1401, opened 2018-06-07
getting name of inner objects adds a $ (GeoShapeQuery)


hoffbeam:
When adding a relation to GeoShapeQuery 
eg. InlineShape(PointShape(GeoPoint(point.lat, point.lon)))).relation(ShapeRelation.within)

the String is outputted as
{"geo_shape":{"areaLocation":{"shape":{"type":"point","coordinates":[15.0,10.0]},"relation":"within$"}}}

the issue is with GeoShapeQueryBodyFn ln 37
q.relation.map(_.getClass.getSimpleName.toLowerCase).foreach(builder.field("relation", _))

Since ShapeRelation.WITHIN is an inner object it is evaluating as "within$"

PR #1402, opened 2018-06-07
Fix relation trailing $ for GeoPointQuery

hoffbeam:
Issue with relation for GeoPointQuery having a trailing '$'

https://github.com/sksamuel/elastic4s/issues/1401


sksamuel:
:+1:

PR #1407, opened 2018-06-08
Change shrink index settings copy builder to use param

tcdevoe:
`ShrinkIndexRequest` was doing a no-op on the `settings` builder (`settings = settings`)


sksamuel:
:+1:

Issue #984, opened 2017-07-08
Refresh query param seems not implemented in delete by query API?


kevinzang:
I wasn't able to specify `refresh` param when calling `delete by query` API, which, according to https://www.elastic.co/guide/en/elasticsearch/reference/5.4/docs-delete-by-query.html, should be supported. I am using v5.4.3 of the HttpClient, and I am working on an sbt project. Here is what I tried:

```
import com.sksamuel.elastic4s.ElasticsearchClientUri
import com.sksamuel.elastic4s.http.HttpClient
import com.sksamuel.elastic4s.http.ElasticDsl._
import org.elasticsearch.action.support.WriteRequest.RefreshPolicy

val client = HttpClient(ElasticsearchClientUri("localhost", 9200))

// First try delete by id. This works as expected. Debug logs show:
// DELETE:/foo/docs/123?refresh=true
client.execute {
  delete("123").from("foo/docs").refresh(RefreshPolicy.IMMEDIATE)
}.await

// Now try delete by query. This does not work as expected. Debug logs show:
// POST:/foo/_delete_by_query?
// {"query":{"range":{"timestamp":{"gte":"2017-06-28T13:07:07.000-07:00"}}}}
// There is no refresh=true in the query param.
client.execute {
  deleteIn("foo").by(rangeQuery("timestamp").gte("2017-06-28T13:07:07.000-07:00")).refresh(RefreshPolicy.IMMEDIATE)
}.await
```

It seems we add refresh param in delete by id case: https://github.com/sksamuel/elastic4s/blob/master/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/delete/DeleteImplicits.scala#L67, but not in delete by query case.

For my use-case, I can simply call `refreshIndex` after `deleteIn`, but I wanted to raise this issue nonetheless in case others have the same issue.

sksamuel:
It's there in the 5.4.x branch. https://github.com/sksamuel/elastic4s/blob/release/5.4.x/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/delete/DeleteImplicits.scala#L66
Can you try the latest 5.4.7 build ?

PR #1408, opened 2018-06-08
Support Refresh of index on delete by query

CarterWilliam:
Thanks for this fantastic library!

A while back this issue was raised: https://github.com/sksamuel/elastic4s/issues/984. It was closed as the delete-by-ID executable already has it, but I think the delete-by-query should have it as well.

The test for delete-by-query was using a 5 second sleep to wait until the elastic index had refreshed but this is no longer necessary when forcing it to refresh on the delete request.


sksamuel:
Excellent thanks. PS, I recently worked with your ex-colleagues at BBC RMS.

PR #1418, opened 2018-06-16
Adds pipeline support to bulk API

kpmeen:
This PR adds pipeline support to the bulk API through `BulkBuilderFn`.


sksamuel:
:+1:

kpmeen:
Thanks for merging this @sksamuel :-)
Do you have any plans for when the next release will be?

sksamuel:
I was keeping master for the 7.0 release, but I might release it as 6.3 as 7.0 is taking a long time to arrive from elastic. If you want it immediately, might be worth doing the same PR on release/6.2.x

kpmeen:
:+1: will do

Issue #1421, opened 2018-06-20
Version Elastic4s 6.29 cannot deserialize Error response and throw an error


t83714:
Version Elastic4s 6.29

Request: ElasticsearchJavaRestClient - Http Response Response{requestLine=GET /regions21/_mapping HTTP/1.1, host=http://localhost:9200, response=HTTP/1.1 404 Not Found}

Error: 

```
com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize instance of `scala.collection.immutable.Map` out of START_ARRAY token
 at [Source: (String)"{"error":{"root_cause":[{"type":"index_not_found_exception","reason":"no such index","resource.type":"index_or_alias","resource.id":"regions21","index_uuid":"_na_","index":"regions21"}],"type":"index_not_found_exception","reason":"no such index","resource.type":"index_or_alias","resource.id":"regions21","index_uuid":"_na_","index":"regions21"},"status":404}"; line: 1, column: 24] (through reference chain: com.fasterxml.jackson.module.scala.deser.MapBuilderWrapper["error"]->com.fasterxml.jackson.module.scala.deser.MapBuilderWrapper["root_cause"])
```
Actual JSON response from elastic search:
```
{
	"error": {
		"root_cause": [
			{
				"type": "index_not_found_exception",
				"reason": "no such index",
				"resource.type": "index_or_alias",
				"resource.id": "regions21",
				"index_uuid": "_na_",
				"index": "regions21"
			}
		],
		"type": "index_not_found_exception",
		"reason": "no such index",
		"resource.type": "index_or_alias",
		"resource.id": "regions21",
		"index_uuid": "_na_",
		"index": "regions21"
	},
	"status": 404
}
```

sksamuel:
Fixed in 6.3.0

PR #1422, opened 2018-06-21
Issue/1421 Fixed GetMappingHandler treat `Index not found` Error as a successful response

t83714:
GetMappingHandler should deal with the `index not found` error probably rather than treat it as a successful response.

Currently, it will treat all response as successful ones and try to convert the JSON response to `Seq[IndexMappings]`.

See issue #1421 

* This fix is for Master branch. *
v6.2.9 needs to update the file `MappingExecutables.scala` with similar logic.


sksamuel:
Thank you.

PR #1432, opened 2018-06-28
feat: add index_options for text fields

michivi:
Added support for the `index_options` mapping parameter for text fields:
https://www.elastic.co/guide/en/elasticsearch/reference/master/index-options.html


sksamuel:
This has been included in 6.3.0

PR #1437, opened 2018-07-03
Fix phrase suggestions

Dbzman:
The text field needs to be outside of the "phrase" object. Also there was a spelling error for "max_errors".
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-suggesters-phrase.html


sksamuel:
:+1:

PR #1442, opened 2018-07-04
Support composite aggregations (#1376)

markuspeters:
Further to issue #1376 and previous PR #1377 against 6.2.x, this PR adds support for composite aggregations against master.


sksamuel:
:+1:

PR #1443, opened 2018-07-04
Fix phrase suggestion highlights

Dbzman:
A new release after merge would be nice 😋


sksamuel:
I can't do 3 release a day :)
I've just done 6.3.2 for you after your last PR. I'll do 6.3.3 next week.

Dbzman:
I can understand that. Is there a way to use the current snapshot/latest master then? (I couldn't find it)
Probably it makes sense to have that snapshots published on commit. What do you think? :)

sksamuel:
I need to set travis to do snapshot release on each commit. I'll look into it.

Dbzman:
Sounds cool, thanks! :)

Issue #1444, opened 2018-07-04
SearchHit `.highlight` can be null


yannick-cw:
If I have no highlights in the `SearchHit` and access the `highlight` field, it is null.

Can we make `highlights` map optional or an empty map by default?

[here](https://github.com/sksamuel/elastic4s/blob/65ce0d8b46e45c2ea0ee7d3191fb7a33194b6b00/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/search/responses.scala#L25)

version `6.2.6`

sksamuel:
Fixed in master.

PR #1445, opened 2018-07-05
fix: nullable highlight in search hits

michivi:
I had encountered the nullable highlight issue as specified in #1444. I was thinking of fixing it like this. This would not break backward compatibility, though any existing client checking for nullity will now do it uselessly...


sksamuel:
Great thanks

PR #1448, opened 2018-07-06
Don’t use hardcoded and deprecated value

Dbzman:
This fix will get rid of the log entry on each request:
```
"Deprecated field [inline] used, expected [source] instead" "Fri, 06 Jul 2018 06:58:21 GMT"
```


sksamuel:
Thanks.

Issue #1447, opened 2018-07-06
SearchError not properly filled


Dbzman:
Hey there,
we noticed that (at least) 404 errors result in SearchError objects which are not filled with any useful information (everything is null there). Elasticsearch delivers the correct response as you can see in #1421. We use elastic4s 6.3.2.
Cheers

sksamuel:
They'll need to be changed individually. Which request type in particular
is causing you issues.

Dbzman:
I'll close this as the linked PR should fix this. :)

mhaddon:
@Dbzman @sksamuel I think this is still a problem, I am on 6.3.5. I am getting SearchError(null,null,null,null,null,null) as a response from a multi query search

sksamuel:
Can you paste in the json that is received. You can turn on logging for com.sksamuel.elastic4s to see it. Obfuscate the values if needed.

mhaddon:
@sksamuel 
The actual error message when I run the query on its own is:

> We failed ElasticError(search_phase_execution_exception,all shards failed,None,None,None,List(ElasticError(illegal_argument_exception,failed to find mapper for [coords] for geo distance based sort,None,None,None,null)))

its just it looses this information when I run the query inside a multi {} block

It looks like the problem is what this PR tried to solve

I will see if I can rebuild everything with logging enabled later, then I can see if I can give you more information

sksamuel:
The logging will give the raw json returned which then we can see why it's not unmarshalling properly. It will be a field misnamed somewhere.

mhaddon:
@sksamuel 
> com.sksamuel.elastic4s = "DEBUG"

in my application.conf?

I dont get any logging output

sksamuel:
You need to enable it whatever logging framework you're using. e4s uses SLF4j, so you might have log4j or logback in your build ?

PR #1449, opened 2018-07-06
Fix SearchError content

alex-voigt:
Fixes issue #1447 
The mapper failed because there was an "error" element wrapping the SearchError


sksamuel:
I wasn't aware this was broken? There's tests that check for this in SearchHttpTest. Let me do some more investigation.

sksamuel:
Oh wait, this for multisearch not search.

sksamuel:
I've added a unit test for this as well.
Thanks for the PR.

PR #1450, opened 2018-07-06
Fix encoder property in highlighting options

stijn-starmind:
The `encoder` and `boundary_chars` options for search highlighting were mixed up because of an apparent copy-paste mistake.


sksamuel:
Thanks!

PR #1451, opened 2018-07-06
Add Starmind to list of companies using Elastic4s

stijn-starmind:
Starmind (http://www.starmind.com/) is using Elastic4s in production. Many thanks :).


sksamuel:
:+1:

Issue #1455, opened 2018-07-09
Not found: value HttpClient


jgnieuwhof:
I'm fairly new to Scala, so please forgive me if this should be obvious.

I've tried running the example here: https://github.com/sksamuel/elastic4s/blob/master/samples/elastic4s-http-client-sbt/src/main/scala/com/sksamuel/elastic4s/samples/HttpClientExampleApp.scala, and despite my best efforts continually receive the following compile error:
```
[error] /home/jgnieuwhof/projects/scala-test/src/main/scala/jgnieuwhof/Main.scala:13:16: not found: value HttpClient
[error]   val client = HttpClient(ElasticsearchClientUri("localhost", 9200))
[error]                ^
[error] one error found
[error] (Compile / compileIncremental) Compilation failed
```

I've added the following to my `build.sbt` file:
```
    libraryDependencies ++= Vector(
      ...,
      "com.sksamuel.elastic4s"      %%  "elastic4s-core"  % "6.3.1",
      "com.sksamuel.elastic4s"      %%  "elastic4s-http"  % "6.3.1"
    )
```

Any ideas? Thanks in advance for any help.

ches:
There have been a lot of changes recently and the documentation is lagging behind, and apparently samples are broken too… Your best bet is to use tests in the elastic4s-tests subproject as a reference, for instance:

https://github.com/sksamuel/elastic4s/blob/master/elastic4s-tests/src/test/scala/com/sksamuel/elastic4s/search/aggs/TermsAggregationHttpTest.scala

You can see [the DockerTests suite mixin][1] for how an HTTP client instance is initialized:

```scala
import com.sksamuel.elastic4s.http.{ElasticClient, ElasticProperties}

val client = ElasticClient(ElasticProperties("http://localhost:9200"))
```

The doc site at https://sksamuel.github.io/elastic4s/docs/ is even more outdated than samples or README I believe, so don't even bother looking there… stick to the tests. I'm going to try to help out with sending some doc update PRs, because I'm going through similar pain of working through misinformation.

[1]: https://github.com/sksamuel/elastic4s/blob/022b56d01d37945ff6b697b33ec1567732aae33b/elastic4s-testkit/src/main/scala/com/sksamuel/elastic4s/testkit/DockerTests.scala

jgnieuwhof:
[...]

PR #1456, opened 2018-07-11
Stopgap Doc Updates

ches:
In light of #1455:

- Removed the samples based on `TcpClient`, since it spews deprecation warnings now I think there's no sense in continuing to show and maintain documentation for it.
- Updated the HTTP client samples to current 6.3.3 APIs.
- Partially updated outdated example code in the README. I could continue on this (particularly the "Connecting to a Cluster" section needs rewriting, probably `TcpClient` should be dropped from there as well), but I would rather invest in reviving tut docs if I can find more time. Let's get examples compiling with the build so they can't go stale again.


sksamuel:
Excellent work.

Touzii:
Hello,
Could we have an example of creating an index using the latest version of elastic4s.
This is my code so far : 

`
def setupIndex(indexName: String) = {
    httpClient.execute {
      createIndex(indexName).alias(inProgressIndexAlias)
    }.flatMap{
      case res: RequestSuccess[CreateIndexResponse] => Future.successful( res.result )
      case err: RequestFailure => Future.failed(new Throwable(err.error.reason))
    }
  }
`

I have the following error : 
>  could not find implicit value for parameter handler: com.sksamuel.elastic4s.http.Handler[com.sksamuel.elastic4s.indexes.CreateIndexRequest,U]

How could I define the implicit Handler.
Thank you guys for considering my request.

PR #1463, opened 2018-07-23
Fix deprecated to point to actual replacement

yannick-cw:
I think `ElasticNodeUri` does not exist and `ElasticNodeEndpoint` is meant here.


sksamuel:
Yes it was, that was an earlier name for it oops.

yannick-cw:
:+1:

PR #1464, opened 2018-07-23
Search request builder won't set 'version' if not defined in request

### LINKED ISSUE #1462: SearchRequest 'version' set to true if not defined by the user
When building a `SearchRequest`, I would assume that if I don't explicitly set the `version` parameter (leaving it to the default of `None`) it shouldn't be included in the request sent to Elasticsearch.

Current behaviour is to set it to true by default 
https://github.com/sksamuel/elastic4s/blob/ac137de93cf301f517c39f913afcd28338ed5d5b/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/search/SearchBodyBuilderFn.scala#L22
and eventually override that by the user provided value, if any.
https://github.com/sksamuel/elastic4s/blob/ac137de93cf301f517c39f913afcd28338ed5d5b/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/search/SearchBodyBuilderFn.scala#L25

Is this on purpose ?

ali-h:
Fixes #1462

### DIFF
--- a/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/search/SearchBodyBuilderFn.scala
+++ b/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/search/SearchBodyBuilderFn.scala
@@ -19,7 +19,6 @@ object SearchBodyBuilderFn {
 
     val builder = XContentFactory.jsonBuilder()
 
-    builder.field("version", true)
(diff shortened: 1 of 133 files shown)


sksamuel:
Tests are both red. Once green I'll merge in.

sksamuel:
:+1:

Issue #1465, opened 2018-07-23
UpdatebyQuery missing scroll


asedghi:
UpdateByQuery is missing scroll on search.  This can cause timeouts on larger result sets.

asedghi:
Ill take care of this later today

PR #1466, opened 2018-07-24
Added scroll to UpdateByQuery

asedghi:
Added scroll for UpdateByQuery.  This is the timeout for the search. Default is 5 min.  
Issue #1465


sksamuel:
:+1:

PR #1468, opened 2018-07-27
[BE-152] Drop unused modules

sophiecollard:
Dropping all modules we don't need. Tests won't compile but we don't care about this.


sophiecollard:
Whoops. Wrong repository :|

sksamuel:
That made me chuckle :)

PR #1473, opened 2018-08-01
Added gzip decompression support to elastic4s-http

### LINKED ISSUE #1472: Gzip compression decoding is not supported
Hi!
I'm trying to enable compression for elastic responses. Elastic supports it, so I added `Accept-Encoding: gzip` header, but when I sending a query, I get an exception: 

`com.sksamuel.elastic4s.http.JavaClientExceptionWrapper: java.nio.charset.UnsupportedCharsetException: gzip`

Looks like it happens because of old unresolved Elastic issue: https://github.com/elastic/elasticsearch/issues/24349, so elastic4s could support gzip decoding itseft.

ilyshav:
Resolves #1472

### DIFF
--- a/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/ElasticsearchJavaRestClient.scala
+++ b/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/ElasticsearchJavaRestClient.scala
@@ -1,6 +1,8 @@
 package com.sksamuel.elastic4s.http
 
+import java.io.InputStream
 import java.nio.charset.Charset
+import java.util.zip.GZIPInputStream
 
 import com.sksamuel.elastic4s.Show
 import org.apache.http.client.config.RequestConfig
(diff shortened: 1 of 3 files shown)


sksamuel:
:+1:

yanns:
Any documentation on how to activate this?

ilyshav:
@yanns You have to add `Accept-Encoding: gzip` header.
I did it with `httpClientConfigCallback`:

```
import scala.collection.JavaConverters._

RestClient
      .builder(new HttpHost("localhost", 9200))
      .setHttpClientConfigCallback((httpClientBuilder: HttpAsyncClientBuilder) => {
        httpClientBuilder.setDefaultHeaders(List(new BasicHeader("Accept-Encoding", "gzip")).asJava)
      }).build
```

Also gzip compression must be enabled in Elastic config:
`http.compression: true `

ilyshav:
> Should we consider activating that per default?

I think we should not enable gzip by default, because this give some profit only is some cases - when Elastic's responses are huge or network perfomance isn't good, but when responses are small, compression will only waste CPU resources , as far as I know. 

> It will only be used the elastic config enables that, right?

I think if Elastic compression will be disabled and we pass accept-encoding header, Elastic will just send plain text responses, so it will work anyway, but this is only my assumption, I'm not checked it.

PR #1482, opened 2018-08-08
Add percolate query support to 5.4.x HTTP client

tarao:
[`percolateQuery` DSL is defined in the core module](https://github.com/sksamuel/elastic4s/blob/62a5cb2bf24c421f13366f3cdf2015068ad82b57/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/searches/QueryApi.scala#L195) but a request body encoding for the query was missing in the HTTP client.

Note that I didn't add [`register` DSL defined in TCP version](https://github.com/sksamuel/elastic4s/blob/62a5cb2bf24c421f13366f3cdf2015068ad82b57/elastic4s-tcp/src/main/scala/com/sksamuel/elastic4s/PercolateDsl.scala#L10) because registering a percolator is just inserting a query to the index.  It is possible by [index API](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html) as the following.

```scala
import com.sksamuel.elastic4s.http.ElasticDsl._
import com.sksamuel.elastic4s.searches.queries.QueryDefinition
import com.sksamuel.elastic4s.http.search.queries.QueryBuilderFn

val query: QueryDefinition = ...
val src = s"""{"query":${QueryBuilderFn(query).string()}}"""
indexInto("some-index" / "percolator-type")
  .source(src)
  .withId("percolator-id")
```

----

I would like to add the same support for other versions than 5.4.x as soon as we move our Elasticsearch clusters to newer versions.


sksamuel:
:+1:

PR #1483, opened 2018-08-08
Add matched_queries support to 5.4.x HTTP client

tarao:
Accept `matched_queries` (a response field for [named queries](https://www.elastic.co/guide/en/elasticsearch/reference/5.4/search-request-named-queries-and-filters.html)) as a field of search hits.

----

I would like to add the same support for other versions than 5.4.x as soon as we move our Elasticsearch clusters to newer versions.


sksamuel:
:+1:

Issue #1458, opened 2018-07-16
Significant terms aggregation result is missing in elastic4s-http


maxcom:
Significant terms aggregation result is missing in elastic4s http client:

https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-significantterms-aggregation.html

sksamuel:
Do you want to add a PR ?

sksamuel:
Addressed by https://github.com/sksamuel/elastic4s/pull/1485

PR #1485, opened 2018-08-08
support for significant terms aggregation result

deepakmohanakrishnan1984:
This PR contains the change to build Significant Terms Aggregation Result.

Addresses the issue reported in #1458


sksamuel:
:+1:

Issue #1391, opened 2018-05-23
Add DocumentExistsDefinition


unkarjedy:
Add convenient API to check whether a document exists like it is done for indexes & types (in addition to IndexExistsDefinition & TypeExistsDefinition)
https://www.elastic.co/guide/en/elasticsearch/guide/current/doc-exists.html

PR #1488, opened 2018-08-13
#1391 check document exists endpoint

unkarjedy:
1) Make index/type/document exists request async, like all the other requests
2) For DocumentExistsHttpExecutable encode document id using URLEncoder. Note that ES doc id can contain any symbol, include url-special.


sksamuel:
:+1:

PR #1491, opened 2018-08-15
Add missing fields to DirectGenerator

TactiC:
We need to be able to set the suggest_mode for a phrase suggestor. And since there were more fields missing I took the liberty to add them all.


sksamuel:
:+1:

Issue #1461, opened 2018-07-20
TermsSetQuery not working inside bool query


cristinaluengoagullo:
Hi,

First of all thanks for the great work you've done with this library! I'm using the release 6.x and I just started using the termsSetQuery function, but I get errors when I try to use it inside a must, filter, etc. I saw that it doesn't inherit from QueryDefinition, could that be the case why it doesn't let me use it inside a must for instance?
Thanks!

cristinaluengoagullo:
Any insight on this please?

sksamuel:
Yep, it should extend 'Query'. I've pushed a change, will do a release soon. This is the 6.3.x release.

sksamuel:
Try 6.3.6

cristinaluengoagullo:
Ok thank you very much!! Once it's available in mvn repository I'll try it.

sksamuel:
It's there now.
http://repo1.maven.org/maven2/com/sksamuel/elastic4s/elastic4s-http_2.12/

Those UI sites take a lot longer to update than the actual repository.

cristinaluengoagullo:
Hi @sksamuel,

Thanks for the very fast reply! Unfortunately I wasn't able to make the query work :( I first tried with the api (termsSetQuery), which accepts the field name and the terms to look for. It returned me the following error:

"caused_by": {
            "type": "illegal_state_exception",
            "reason": "No minimum should match has been specified"
          }

Then I tried building the query without the api, calling TermsSetQuery with minimumShouldMatchScript and a very simple script, but the resulting query didn't have the minimum_should_match_script field. I checked the code, and perhaps it has to do with this (if not, please let me know!):

object TermsSetQueryBodyFn {
  def apply(t: TermsSetQuery): XContentBuilder = {
    val builder = XContentFactory.jsonBuilder().startObject("terms_set")
    builder.startObject(t.field)
    **builder.array("terms", t.terms.map(_.toString).toArray[String])
    t.minimumShouldMatchField.foreach(builder.field("minimum_should_match_field", _))**
    builder.endObject().endObject()
  }
}

It seems to build the "minimum_should_match_field" if present, but it doesn't add the "minimum_should_match_script" field. Then I tried setting the "minimumShouldMatchField", but since it only accepts an Option[Int], I can't make it point to any field in the documents (for what I understand, the minimum_should_match_field's value should be the name of a field). 

Again, thank you so much for your work and diligence! And if there's anything wrong in what I said please let me know!

[...]

PR #1505, opened 2018-08-22
Terms in set query support

cristinaluengoagullo:
This is regarding issue #1461

### DIFF
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/searches/QueryApi.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/searches/QueryApi.scala
@@ -258,11 +258,16 @@ trait QueryApi {
   def termsLookupQuery(field: String, path: String, ref: DocumentRef) =
     TermsLookupQuery(field, TermsLookup(ref, path))
 
+  // Either minimumShouldMatchField or minimumShouldMatchScript should be specified, that's why they appear as mandatory parameters
   def termsSetQuery(field: String, 
-      terms: Set[Any], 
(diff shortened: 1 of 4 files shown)


elastic4s-core/src/main/scala/com/sksamuel/elastic4s/searches/QueryApi.scala
```diff
@@ -258,8 +258,11 @@ trait QueryApi {
   def termsLookupQuery(field: String, path: String, ref: DocumentRef) =
     TermsLookupQuery(field, TermsLookup(ref, path))
 
-  def termsSetQuery(field: String, terms: Set[Any]): com.sksamuel.elastic4s.searches.queries.term.TermsSetQuery =
-    com.sksamuel.elastic4s.searches.queries.term.TermsSetQuery(field, terms)
+  def termsSetQuery(field: String,
```
sksamuel: Since one of the two minimumShouldMatch params have to be specified, I think we should have two methods here, one that takes a string, and one that takes a script. Then we can do,

`termsSetQuery(field, terms, min)` or `termsSetQuery(field, terms, script("script"))` rather than `termsSetQuery(field, terms, Some(min))`. If that's ok, please make the change and I'll merge.

Great pr thanks.

cristinaluengoagullo: True! Might be clearer when using the method. I'll make the change, thanks!

sksamuel:
:+1:

cristinaluengoagullo:
Thanks :)!

Issue #1314, opened 2018-03-22
ScriptedMetricAggregationDefinition not supported


therealnb:
Hello;
I'm trying to use `ScriptedMetricAggregationDefinition` with elastic4s-http 5.4.15. I get the following exception
```
scala.MatchError: ScriptedMetricAggregationDefinition(ids,Some(ScriptDefinition(params._agg.ids = [],Some(painless),inline,Map(),Map())),Some(ScriptDefinition(params._agg.ids.add(doc._uid.value),Some(painless),inline,Map(),Map())),Some(ScriptDefinition(,Some(painless),inline,Map(),Map())),Some(ScriptDefinition(def tot = new HashSet(); for (a in params._aggs) { tot.addAll(a.ids) } return tot,Some(painless),inline,Map(),Map())),Map(),List(),Map()) (of class com.sksamuel.elastic4s.searches.aggs.ScriptedMetricAggregationDefinition)
	at com.sksamuel.elastic4s.http.search.aggs.AggregationBuilderFn$.apply(AggregationBuilderFn.scala:30)
	at com.sksamuel.elastic4s.http.search.SearchBodyBuilderFn$.$anonfun$apply$63(SearchBodyBuilderFn.scala:204)
	at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:59)
	at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:52)
	at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:48)
```
which I think I understand as a missing case in `AggregationBuilderFn`.

I posted this question [https://stackoverflow.com/questions/49428759/tweaking-elastic4s-queries-before-they-are-sent](https://stackoverflow.com/questions/49428759/tweaking-elastic4s-queries-before-they-are-sent) to ask for a workaround.

The aim is to use this aggregation, which doesn't seem directly possible in this branch, but the question here is 'is there a way to add this manually' before query?

I have a bunch of other aggregations and parameters defined in the DSL, and it seems to me I have to throw all that out and do a raw query. I don't really want to do that. I would like to create the query using the DSL, then change a bit manually, then send.

therealnb:
I got round this in my code by subclassing AggregationBuilderFn and providing ScriptedMetricAggregationBuilder. This is very hacky, so I am open to better suggestions.

For reference, if anyone else want to hack in the same way.

```
package com.sksamuel.elastic4s.http.search.aggs

[...]

PR #1507, opened 2018-08-22
Added ScriptedMetricAggregationBuilder to the HTTP DSL for 5.4.x

paulporu:
Same code as #1506, might as well use it to tackle #1314


sksamuel:
:+1:

PR #1509, opened 2018-08-23
Only generate collate if needed

TactiC:
When not providing values for the collate function an empty block gets created like this:
"collate": {
  "query":{}
}
Which results in deprecation errors: _must specify either [source] for an inline script or [id] for a stored script_
This pr omits the collate block if not explicitly needed.


sksamuel:
Can you fix the failing test?

sksamuel:
Thanks

PR #1511, opened 2018-08-23
Support HighlightOptions at the level of a nested query

stijn-starmind:
* Support using `HighlightOptions` at the level of a nested query.
* Unify how highlighting is defined on a `SearchRequest` and on an `InnerHit` respectively.


sksamuel:
awesome

PR #1514, opened 2018-08-27
Added average, cardinality, and filters AggregationResult

paulporu:
The purpose of this pull request is to add the "average", "cardinality", and "filters" AggregationResult objects to the HTTP DSL in order to extend the current features for parsing the SearchResponse. However, I came across two issues during the implementation: 

1. The already implemented `Bucket` object possessed the "key" and "doc_count" properties but did not allow retrieving the actual content of the bucket
2. The types of all "doc_count", "score", and "total_hits" properties were not consistent with the types defined by the [Elasticsearch Java API](https://github.com/elastic/elasticsearch/blob/v5.6.2/core/src/main/java/org/elasticsearch/search/aggregations)

I resolved the first issue by adding an `agg_data` field to the `Bucket` object to allow the parsing of all the content of the bucket. I resolved the second issue by changing those types according to the type definitions declared in the Elasticsearch Java API. Both changes are breaking compatibility with previous versions of this 5.6.x branch.
I realize this is a very opinionated approach but I figured those were issues waiting to be discovered rather than features of the provided API and so decided to go ahead and fix them at the same time as adding the new AggregationResult objects. I come with an open mind and full transparency on this and await your feedback so we can best figure out how to move forward.

Best,
Paul.


sksamuel:
Thanks for the PR. I think changing the ints to longs is fine, but I don't think the doubles need to become floats. All we're doing there is breaking the API without adding any functionality or fixing anything, since a float value can always be held in a double value.

paulporu:
That's totally fair. I've reverted this change and updated the PR accordingly.

paulporu:
@sksamuel Thanks for the quick merging. Would you mind releasing version 5.6.7 that would include all the latest additions?

sksamuel:
Done.

paulporu:
Perfect, cheers

PR #1516, opened 2018-08-31
Debug ScoreMode.scala

eramonk:
ScoreMode.Total is a Lucene constant, it corresponds to sum score mode in Elasticsearch documentation and HTTP API.

https://github.com/elastic/elasticsearch/blob/20e1b5e2fe3e7319e2c908657e484bb49bbf11d7/server/src/main/java/org/elasticsearch/index/query/NestedQueryBuilder.java#L229


sksamuel:
:+1:

PR #1517, opened 2018-09-05
fix a warning when using RandomScoreFunction

cpestour:
fix the warning "As of version 7.0 Elasticsearch will require that a [field] parameter is provided when a [seed] is set"
when using RandomScoreFunction


sksamuel:
:+1:

PR #1528, opened 2018-09-20
Add/improve top hits aggregation response, fix histogram

yohannj:
Regarding top hits, replaced some _.asInstanceOf[X]_ by _.toString.toX_ in order to avoid some ClassCastException from Int to Long for example. That's what is already used in other methods.

Also, according to top hits aggregation documentation: https://www.elastic.co/guide/en/elasticsearch/reference/5.6/search-aggregations-metrics-top-hits-aggregation.html
_score and max_score can be null, so I switched to _Option[Double]_

Regarding the histogram aggregation response, this is a side fix as the return response is a Seq of Map, not just a Map.


sksamuel:
Would be good to have the changes on master too.

PR #1534, opened 2018-10-02
Update json libraries in documentation

tuleism:
As the title. I also update `tut`, but it doesn't seems to be in used anymore?


sksamuel:
Yeah I don't use it.

Issue #1535, opened 2018-10-04
Illegal character error while using 'near' operator in search query (5.5.11)


SovietSir:
Hello,

Recently I have faced with such issue - when I try to perform simple search request with 'near' operator, I get following error:

> Exception in thread "main" org.elasticsearch.client.ResponseException: POST http://localhost:9200/index/type/_search: HTTP/1.1 500 Internal Server Error
{"error":{"root_cause":[{"type":"json_parse_exception","reason":"Illegal character ((CTRL-CHAR, code 0)): only regular white space (\\r, \\n, \\t) is allowed between tokens\n at [Source: org.elasticsearch.transport.netty4.ByteBufStreamInput@1cc2525e; line: 1, column: 90]"}],"type":"json_parse_exception","reason":"Illegal character ((CTRL-CHAR, code 0)): only regular white space (\\r, \\n, \\t) is allowed between tokens\n at [Source: org.elasticsearch.transport.netty4.ByteBufStreamInput@1cc2525e; line: 1, column: 90]"},"status":500}
	at org.elasticsearch.client.RestClient$1.completed(RestClient.java:354)
	at org.elasticsearch.client.RestClient$1.completed(RestClient.java:343)
	at org.apache.http.concurrent.BasicFuture.completed(BasicFuture.java:119)
	at org.apache.http.impl.nio.client.DefaultClientExchangeHandlerImpl.responseCompleted(DefaultClientExchangeHandlerImpl.java:177)
	at org.apache.http.nio.protocol.HttpAsyncRequestExecutor.processResponse(HttpAsyncRequestExecutor.java:436)
	at org.apache.http.nio.protocol.HttpAsyncRequestExecutor.inputReady(HttpAsyncRequestExecutor.java:326)
	at org.apache.http.impl.nio.DefaultNHttpClientConnection.consumeInput(DefaultNHttpClientConnection.java:265)
	at org.apache.http.impl.nio.client.InternalIODispatch.onInputReady(InternalIODispatch.java:81)
	at org.apache.http.impl.nio.client.InternalIODispatch.onInputReady(InternalIODispatch.java:39)
	at org.apache.http.impl.nio.reactor.AbstractIODispatch.inputReady(AbstractIODispatch.java:114)
	at org.apache.http.impl.nio.reactor.BaseIOReactor.readable(BaseIOReactor.java:162)
	at org.apache.http.impl.nio.reactor.AbstractIOReactor.processEvent(AbstractIOReactor.java:337)
[...]

PR #1537, opened 2018-10-08
Fixed bug with extra zero bytes

unkarjedy:
Fixed bug with extra zero bytes, simplified logic of building array fields (for near queries), removed low level code, added critical tests for XContentBuilderExtensions

refers to https://github.com/sksamuel/elastic4s/issues/1535


sksamuel:
:+1:

PR #1538, opened 2018-10-09
Fix travis build for scala 2.12 on 1.5.x branch

bcolyn:
Heya, I'm a colleague of @sebastiaansamyn-tc and this is a follow-up on #1381 - it seems that building the 1.5 branch for scala 2.12 on travis is failing. There was no config for cross-compiling in .travis.yml and it should build with JDK8, not 7. I've also bumped scalaversion to 2.12.7 from 2.12.6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,11 @@
 language: scala
 scala:
-- 2.11.6
+- 2.11.12
+- 2.12.7
 script:
 - sbt clean test
 jdk:
-- oraclejdk7
-- openjdk7
+- oraclejdk8
+- openjdk8
 env:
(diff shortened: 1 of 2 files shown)


bcolyn:
Hopefully, this PR would result in binary artifacts of 1.5.x for 2.12

sksamuel:
`sbt "release skip-tests"` does not want to do a 2.12 release. If you can fix that I'll do a release. @bcolyn

bcolyn:
@sksamuel i think you need sbt "+release skip-tests" to make sure all cross-build versions are published

does that solve your issue?

sksamuel:
Usually you don't need to do +release as there's an sbt key that sets cross releases. I'll try it anyway.

sksamuel:
Ok I tried `sbt "+release skip-tests"` which has the weird effect of _only_ publishing for 2.12. I've copied some settings from master and it works with `sbt "release skip-tests"` as master does now. 1.5.22 released.

sebastiaansamyn-tc:
Thanks, this is much appreciated!

PR #1539, opened 2018-10-12
Added Elasticsearch common analyzers to elastic4s-embedded local node

kraanerk:
Elasticsearch built-in analyzers were moved into an Elasticsearch plugin: analysis-common.
If this plugin is not registered by LocalNode testing indexes using built-in analyzers with elastic4s-embedded is not possible.


sksamuel:
:+1:

PR #1543, opened 2018-10-18
Fix akka-http client compilation for Scala 2.11

DGolubets:
Sorry, I forgot about cross-compilation.


sksamuel:
I don't think putting the scala version in travis actually does anything.

PR #1545, opened 2018-10-22
Fix BulkCompatibleRequest type in readme

giftig:
The code snippet was wrong; the stated type doesn't exist and should
actually be called `BulkCompatibleRequest` rather than
`BulkCompatibleDefinition`.


sksamuel:
:+1:

PR #1546, opened 2018-10-24
Fixed bug with extra zero bytes

unkarjedy:
Fixed bug with extra zero bytes, simplified logic of building array f…elds (for near queries), removed low level code, added critical tests for XContentBuilderExtensions


unkarjedy:
This is a copy of https://github.com/sksamuel/elastic4s/pull/1537 for 5.6.x version

Issue #1547, opened 2018-10-25
Use scala-loggin all over the project


unkarjedy:
While Elastic4s  is a syntax-sugar around ES Java native client (as I see it) I think we could use the power of Scala scala-logging library: https://github.com/lightbend/scala-logging

It goes without saying that Elastic4s has tons of places with an unoptimized usage of Log4j logger:
1) Using interpolation instead of logger method parameters
2) Not checking whether some log level is enabled
This leads us to tons of **really** unnecessary allocations and heavy usage of JVM GC. 

Scala-Loggin library is quite lightweight wrapper around Log4j but it is very useful.

I can prepare a PR.

PR #1548, opened 2018-10-25
Optimized http bulk api by lazy logging and eliminating unnecessary memory allocations abd GC cycles

unkarjedy:
In the current implementation of HTTP bulk API ther is quite a havy and unoptimized place in class
```com.sksamuel.elastic4s.http.bulk.BulkImplicits.BulkExecutable```:
```scala
val rows = BulkContentBuilder(bulk)
logger.debug(s"Bulk entity: ${rows.mkString("\n")}")
// es seems to require a trailing new line as well
val entity = new StringEntity(rows.mkString("\n") + "\n", ContentType.APPLICATION_JSON)
```
I would like to point out that bulk API should be considered with a certain care since we talk about `O(N)` memory usage.

Let's say that we have a bulk of N elements and the total size of JSONs is ~100Mb.
(The question of what bulk size  in elements/bytes is out of scope).
```BulkDefinition.requests``` will contain those ~100Mb bulk (eg. in ```IndexDefinition.source``` fields)
```val rows = BulkContentBuilder(bulk)``` will also contain ~100Mb rows
first ```rows.mkString("\n")``` (during logging) will allocate another ~100Mb.
When it is interpolated inside the string (during logging) it will allocate another ~100Mb
line ```new StringEntity(rows.mkString("\n") + "\n"``` will require extra 3 * 100Mb = 300mb (including mem copy inside StringEntity)

more of that when logging huge body inside:
```client.async("POST", endpoint, params.toMap, entity, ResponseHandler.default)```
```logger.debug(Source.fromInputStream(entity.getContent)(Codec.UTF8).getLines().mkString("\n"))```
it allocates another ~200Mb for `getLines` and `mkString`

So when indexing a single bulk just this simple place of code uses ~800Mb (**x8**) extra memory out of nowhere even if we do not use debug level for elastic.

I understand that most of the memory is ready for garbage collection quite fast but it is a really unnecessary load for GC that we could avoid.

This PR reduces extra memory usage to **x3** extra memory in non-debug mode **x4** in debug mode.

also relates to:  https://github.com/sksamuel/elastic4s/issues/1547

### DIFF
--- a/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/HttpExecutable.scala
+++ b/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/HttpExecutable.scala
@@ -43,9 +43,16 @@ trait HttpExecutable[T, U] extends Logging {
               endpoint: String,
               params: Map[String, Any],
               entity: HttpEntity,
-              handler: ResponseHandler[U]): Future[U] = {
-      logger.debug(s"Executing elastic request $method:$endpoint?${params.map { case (k, v) => k + "=" + v }.mkString("&")}")
-      logger.debug(Source.fromInputStream(entity.getContent)(Codec.UTF8).getLines().mkString("\n"))
+              handler: ResponseHandler[U],
+              skipBodyLogging: Boolean = false): Future[U] = {
+      if (logger.isDebugEnabled()) {
+        logger.debug(s"Executing elastic request $method:$endpoint?${params.map { case (k, v) => k + "=" + v }.mkString("&")}")
+        val bodyLogged =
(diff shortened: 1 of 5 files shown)


unkarjedy:
As for ```skipBodyLogging: Boolean = false``` I do not think that we should log bulk body inside HttpExecutable.async because we already do that inside ```BulkExecutable``` and the body can be really huge.

unkarjedy:
@sksamuel

sksamuel:
Looks great.

sksamuel:
The same change on master would be awesome.

unkarjedy:
done for master: https://github.com/sksamuel/elastic4s/pull/1552

PR #1553, opened 2018-10-30
Optimized http bulk api by lazy logging and eliminating unnecessary memory allocations abd GC cycles

unkarjedy:
copy of: https://github.com/sksamuel/elastic4s/pull/1548


sksamuel:
Thank you.

PR #1558, opened 2018-11-07
Fix synonymTokenFilter method

danslapman:
SynonymTokenFilter contains require which fails when constructor is called with empty path & synonyms.


sksamuel:
:+1:

PR #1562, opened 2018-11-13
Add instances for cats-effect IO, Monix Task and Scalaz Task

agustafson:
There's some good work that has been done to create modules and executors for a cats-effect `IO`, Monix `Task` and scalaz `Task, but there are a couple of things which are missing thus far, namely an instance of `Functor[F]` and the lack of a single import which can provide implicit instances of the `Functor` and `Executor`.

I've included simple `instances` imports and updated the readme.

Happy to change the import names or structure if you'd prefer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,8 @@ To get started you will need to add a dependency to either
 depending on which client you intend you use (or both).
 
 The basic usage is that you create an instance of a client and then invoke the `execute` method with the requests you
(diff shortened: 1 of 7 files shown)


sksamuel:
Excellent work.

agustafson:
Holy crap that was quick! Thanks for the great work with elastic4s (and other projects) @sksamuel 👍

sksamuel:
You're welcome :)

PR #1564, opened 2018-11-19
Parameter time_zone not set in date_range aggregation

lmafonso:
This PR sets parameter timeZone in date_range aggregation, since it has been forgotten. 
Thanks.


sksamuel:
:+1:

sksamuel:
Thanks, I have applied this change manually.

PR #1567, opened 2018-11-26
Fs2 search result subscriber

agustafson:
This adds a search result subscriber which returns an `fs2.Stream` of `SearchHit`s. It's similar or an alternative to the reactive streams subscription code, but uses fs2 rather than akka to manage the streaming of results.

I'll add some details to the README too if this seems like the code is ok.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -21,6 +21,7 @@ lazy val root = Project("elastic4s", file("."))
     sttp,
     akka,
     httpstreams,
+    fs2streams,
     embedded
   )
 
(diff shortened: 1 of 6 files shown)


sksamuel:
I appreciate the PR but I feel that a FS2 module would be better serviced as a standalone project.
The reason are that I don't use it myself and so won't have time or knowledge to maintain it when the FS2 API changes or bug reports are raised.

agustafson:
Fair enough @sksamuel - thanks for the feedback. I'd be happy to maintain it if need be, but equally ok with extracting it into another project.

sksamuel:
I'd love to have it part of the project, but what can happen is the original author is too busy with real life, and bug requests get filed and I have to fix them, and I don't have the knowledge for fs2 fixes. I'm more an akka-streams man. So I have to say that only stuff that I can maintain myself if needs be can be merged in. So I love the PR and the effort you've put in, but I have to graciously decline if that's ok.

PR #1568, opened 2018-11-26
Operations on /_cat/*  don’t work with elastic4s-akka client

lustefaniak:
This is caused by `uri = Uri(request.endpoint).withQuery(Query(request.params)),` code in `AkkaHttpClient.toRequest` which always overrides all query string parameters passed as part of endpoint parameter with empty `request.params` map.


sksamuel:
Excellent!

PR #1570, opened 2018-11-29
Add failover to akka-http client

DGolubets:
I've tried to implement here similar retry and host blacklisting logic as in the official REST client:
1. Exceptions and some HTTP status codes will trigger a retry until ```maxRetryTimeout``` is reached.
2. Exceptions and some HTTP status codes will mark a host dead and it will be excluded from round-robin for some time starting with ```blacklistMinDuration``` and up to ```blacklistMaxDuration```.

I've done some refactoring as well, primarily to make the client more unit-test friendly.


sksamuel:
Fantastic!

PR #1571, opened 2018-11-30
Small Bugfix on 5.6.X TCP Branch

jcern:
I am not sure your policy on pull requests, but this pull fixes two small issues that we had encountered in production. I know this is an older branch, but I wanted to submit it in case it would be useful to anyone else. 

The first removes a println statement that appears to leftover debugging info, and can't be disabled otherwise.

The second allows for the `ignoreUnmapped` parameter to be passed into the builder, currently - the parameter is ignored, and therefore generates a query with a value of `false` (since that is the default value in the java client).


sksamuel:
Excellent thanks.

Issue #1573, opened 2018-12-05
Path prefix support does not actually work.


jrouly:
My usage:

```scala
val properties: ElasticProperties = ElasticProperties("http://example.com:80/es/") // not real domain
val client: ElasticClient = ElasticClient(properties)
...
```

Selected stack trace:
```
com.sksamuel.elastic4s.http.JavaClientExceptionWrapper: java.lang.NullPointerException
...
Caused by: java.lang.NullPointerException: null
...
Suppressed: org.elasticsearch.client.ResponseException: method [GET], host [http://example.com:80], URI [/modeled-documents/_count], status line [HTTP/1.1 404 Not Found]
```

Cause: https://github.com/sksamuel/elastic4s/blob/master/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/ElasticClient.scala#L85 `prefix` is ignored here and not passed into the underlying client.

PR #1574, opened 2018-12-05
Fix support for path prefixing.

jrouly:
Addresses https://github.com/sksamuel/elastic4s/issues/1573.

Path prefixs are being read in and then immediately ignored.

I added a (janky) test in `ElasticClientPrefixTests` that goes about testing in a somewhat roundabout fashion. Let me know if you have suggestsions for improving this test.


jrouly:
Hey @sksamuel please review when you have a moment. I've tested this downstream in my application and it seems to do the trick.

jrouly:
@sksamuel Bad things happen.

This test passes currently in `ElasticPropertiesTest`.

```scala
    val properties = ElasticProperties("https://host1:1234/prefix1?foo-bar,host2:1234/prefix2?fizz=buzz")

    properties.prefix.value shouldEqual "/prefix1"
    properties.options shouldEqual Map("foo-bar,host2:1234/prefix2?fizz" -> "buzz")
    properties.endpoints.loneElement.protocol shouldEqual "https"
    properties.endpoints.loneElement.host shouldEqual "host1"
    properties.endpoints.loneElement.port shouldEqual 1234
```

I'll take a look at the regexes, I think they may need some love.

jrouly:
Although to be honest, I would sort of expect this string to fail:

```
https://host1:1234/prefix1?foo-bar,host2:1234/prefix2?fizz=buzz"
```

And this string to pass:

```
https://host1:1234,host2:1234/prefix2?fizz=buzz"
```

that is -- a *host* is only associated with a hostname and a port, so they should be grouped in a comma separated list. The path should be applied to all of the hosts uniformly -- those are the semantics of the ES SDK, at least.

PR #1575, opened 2018-12-05
Add preference support for MultiSearch

Dbzman:
Hey there,

we noticed that the MultiSearch feature didn't support specifying a preference. This should fix it.
I'm not sure if the test is placed in the correct file. You may move it.

Cheers
Timo


sksamuel:
:+1:

PR #1578, opened 2018-12-07
Mirror HitReader typeclass instances with AggReader.

jrouly:
`HitReader` is in all the json packages, but `AggReader` wasn't complete.

The only implementation missing now is the low level Jackson implementation, which is a lot more involved than these easy movers.

@sksamuel


jrouly:
Cherry picked 8b68931 from https://github.com/sksamuel/elastic4s/pull/1574 to address the flaky test.

Issue #1582, opened 2018-12-13
JsonMapperException when parsing incomplete error response 


butleradamj-deere:
We are occasionally seeing errors like these in our logs:  

```
com.fasterxml.jackson.databind.JsonMappingException: No content to map due to end-of-input
at [Source: ; line: 1, column: 1]
at com.fasterxml.jackson.databind.JsonMappingException.from(JsonMappingException.java:148)
at com.fasterxml.jackson.databind.ObjectMapper._initForReading(ObjectMapper.java:3781)
at com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:3721)
at com.fasterxml.jackson.databind.ObjectMapper.readTree(ObjectMapper.java:2271)
at com.sksamuel.elastic4s.http.ElasticError$.parse(ElasticError.scala:34)
at com.sksamuel.elastic4s.http.DefaultResponseHandler.handle(ResponseHandler.scala:49)
at com.sksamuel.elastic4s.http.HttpClient$$anonfun$execute$1.apply(HttpClient.scala:69)
at com.sksamuel.elastic4s.http.HttpClient$$anonfun$execute$1.apply(HttpClient.scala:68)
at scala.util.Success$$anonfun$map$1.apply(Try.scala:237)
at scala.util.Try$.apply(Try.scala:192)
at scala.util.Success.map(Try.scala:237)
at scala.concurrent.Future$$anonfun$map$1.apply(Future.scala:237)
at scala.concurrent.Future$$anonfun$map$1.apply(Future.scala:237)
at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:36)
at scala.concurrent.impl.ExecutionContextImpl$AdaptedForkJoinTask.exec(ExecutionContextImpl.scala:121)
at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
at scala.concurrent.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979)
at scala.concurrent.forkjoin.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:107)
```

This is being thrown here:  https://github.com/sksamuel/elastic4s/blob/master/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/ElasticError.scala#L37

[...]

PR #1583, opened 2018-12-13
Fix for JsonMapperException in ElasticError

butleradamj-deere:
Addresses #1582 

Adds error handling around potential JsonMapperExceptions if the response body in an ElasticSearch error is incomplete or otherwise malformed.


sksamuel:
I've merged this manually, thanks.

Issue #1585, opened 2018-12-19
Missing script handling in percentile aggregation


napgab:
Percentile aggregation works only on a field. 
The script definition is not serialized at all in the PercentilesAggregationBuilder.
https://github.com/sksamuel/elastic4s/blob/master/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/search/aggs/PercentilesAggregationBuilder.scala

The same issue was already submitted past year, resolved and tagged as closed, but seems still unresolved.
https://github.com/sksamuel/elastic4s/issues/948

sksamuel:
If you add a PR I will do a release with the fix thanks.

napgab:
PR created. Because this issue is a bit old, will be great integrate this fix also in older versions, not only on the latest release.

sksamuel:
If there's a version you need the fix releasing against, please make a PR against that version and I will do a release.

PR #1586, opened 2018-12-19
PercentilesAggregationBuilder: handle aggregation with script

napgab:
https://github.com/sksamuel/elastic4s/issues/1585


sksamuel:
Thanks, I have applied this change manually as the modules have changed recently.

PR #1610, opened 2019-01-17
Support geo-location in completion suggester

idiotech:
Elasitcsearch support "geo" and "category" contexts in completion suggester, but elatic4s supports only the latter. This patch adds geo support.


idiotech:
I couldn't add unit tests becasue the whole ```CompletionSuggestionsTest``` class is commented out...

sksamuel:
:+1:

PR #1629, opened 2019-02-04
Fix SearchType for MultiSearch

Dbzman:
Hey again,

here's another fix for MultiSearch. During implementation I noticed that you specified the default search type as `DfsQueryThenFetch`. The official [Elasticsearch documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-search-type.html#query-then-fetch) although states that the default is `QueryThenFetch`. I changed that. There's no test failing except for the `DateRangeQueryHttpTest` which is also broken on master.

It would be cool to get this also into a 6.5 release of elastic4s.

Cheers and many thanks!
Timo


sksamuel:
:+1:

Issue #1646, opened 2019-02-15
realtime, preference, and refresh params not used in mget calls


rengawm:
The ES `mget` API supports passing the `realtime`, `preference`, and `refresh` query parameters just like the `get` API does. (Reference: https://github.com/elastic/elasticsearch/blob/6.7/server/src/main/java/org/elasticsearch/rest/action/document/RestMultiGetAction.java#L62-L64)

These fields exist in elastic4s's `MultiGetDefinition`/`MultiGetRequest` classes but never get put on the resulting HTTP request. It seems like they'd just need to be sent as params here: https://github.com/elastic/elasticsearch/blob/6.7/server/src/main/java/org/elasticsearch/rest/action/document/RestMultiGetAction.java#L62-L64

PR #1647, opened 2019-02-15
Add support for realtime, preference, and refresh params in multiget

rengawm:
This addresses #1646.

A couple things I was not sure about:
- Looking quickly, I didn't see existing examples of ensuring that we generate the expected get requests (including params, specifically), so I haven't added tests for this.
- I'd be happy to add this to the `multiget.md` doc... It seems like it'd have to go into a whole new usage pattern for how to set these fields, though - there doesn't seem to be a nice way to fit them into the `multiget(item, item, item)` DSL shown there, so it would have to show manual construction of the `MultiGetRequest` class.


sksamuel:
Excellent PR thank you.
For docs, if you do `multiget(item, item, item).refresh(value)` etc that should work.

PR #1661, opened 2019-03-07
Don't URL encode index names when used in request body.

bpiper:
Also handle wait_for_completion in RestoreSnapshotRequest.


sksamuel:
:+1:

PR #1663, opened 2019-03-11
Added Utility Function getAgg to `HasAggregations` trait

t83714:
Added Utility Function `getAgg` to `HasAggregations` trait so that it'd easier to traverse cascaded aggregations in a generic way.
e.g.:
```scala
val agg:Option[Aggregations] = response.aggregations.getAgg("named-aggration")
```
or 
```scala
val agg:Option[Aggregations] = response.aggregations
                                          .getAgg("named-aggration")
                                          .flatMap(_.getAgg("sub-named-aggreation"))
```


sksamuel:
:+1:

Issue #1659, opened 2019-03-06
"source" field on SearchHit and InnerHit is Map[String, AnyRef]


edubrovski:
Working with fields of type Map[String, AnyRef] is annoying and not Scala way. It'd be great to have a way to get source as a JSON AST. Ideally, I'd like to do something like this:
`import play.json.api.libs.json._
import com.sksamuel.elastic4s.playjson._
val jsonReads: Reads[MyScalaClass] = ... 
val hit: SearchHit = ...
hit.sourceAsJson // returns JsValue from play-json
hit.sourceAs[MyScalaClass] // returns MyScalaClass by getting JsValue and converting it using defined Reads[MyScalaClass]`

sksamuel:
You can use HitAs for this.

edubrovski:
We can also do the following:
```
val raw = SourceAsContentBuilder(innerHit.source).string() // or just hit.sourceAsString in case of SearchHit
val parsed = Json.parse(raw).as[MyScalaClass]
```
But JSON is parsed twice in this sample.

sksamuel:
So you want to marshall the response from elastic directly into your own type?
If so, you could make an implicit handler and bring that into scope.

edubrovski:
I think I got it. After importing `com.sksamuel.elastic4s.playjson._` I can do `hit.to[MyScalaClass]`. This is great, but only SearchHit has to and safeTo methods. Can we do the same for InnerHit? If not, I can add some methods to InnerHit and create a PR.

sksamuel:
Of course please do.

edubrovski:
I wanted to extend Hit trait with InnerHit and use HitReader typeclass for reading inner hits, but I couldn't because of version and score fields on Hit. So I've added a convenience method for converting Map[String, Any] to json string.

PR #1666, opened 2019-03-12
Add sourceAsString method to InnerHit

edubrovski:
This method is the same as SearchHit#sourceAsString.

Relates to #1659.


sksamuel:
:+1:

PR #1670, opened 2019-03-14
Fix text field index options

ollik1:
TextField contains parameter `index_options` but it is unused in
CommonFieldBuilder omitting the value from mappings.


sksamuel:
:+1:

PR #1674, opened 2019-03-20
Strip dollar sign suffix on range relation name

### LINKED ISSUE #1673: Incorrect relation name
Same issue as #1401  but on range queries.

jbenoit2011:
Fixes #1673 
Based on #1402


sksamuel:
:+1:

PR #1676, opened 2019-03-26
Add level parameter to cluster health query

stringbean:
This adds support for the `level` parameter to the `clusterHealth` request which allows querying health at the index and shard level.

The default `clusterHealth` query will continue to behave as it does currently and adding a call to `level` will request the extra information:

```scala
// default (cluster level)
val health1 = client.execute {
    clusterHealth()
  }.await.result
// health.indices == null

// index level
val health2 = client.execute {
    clusterHealth() level ClusterHealthLevel.Indices
  }.await.result
// health2.indices == Map(index -> health)

// shard level
val health3 = client.execute {
    clusterHealth() level ClusterHealthLevel.Shards
  }.await.result
// health3.indices == Map(index -> health) with health.shards == Map(shard -> health)
```

Also added tests for the existing & new cluster health code.


sksamuel:
:+1:

PR #1680, opened 2019-04-01
use string instead of int for minimum_should_match in query_string and simple_query_string

fabriziofortino:
This is needed to specify absolute values, percentages or a combination of both
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-minimum-should-match.html


sksamuel:
:+1:

PR #1681, opened 2019-04-01
add auto_generate_synonyms_phrase_query parameter

fabriziofortino:
Add to multi_match_query, query_string and simple_query_string.

Introduced in ES 6.1 https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-match-query.html#query-dsl-match-query-synonyms


sksamuel:
Thank you for this. I can't merge because of a conflict, if you could fix that I would merge for the next release.

sksamuel:
:+1:

Issue #948, opened 2017-06-13
Issue when building percentiles aggregation (missing percentils)


pgorin:
PercentilesAggregationBuilder is not setting percentiles when building aggregation.
file: com.sksamuel.elastic4s.searches.aggs.PercentilesAggregationBuilder

sksamuel:
Fixed in 6.0.0-rc1

pgorin:
Thanks!

pgorin:
Hi Stephen, how are you. I have a question. I'm currently using elastic4s version 5.4.5 over elasticsearch version 5.5. I see that you fixed this bug on elastic4s 6.x versions. Is there a chance that you can include this fix on elastic4s 5.5.x version? Thank!

PR #1682, opened 2019-04-02
#948 missing percentiles fix into 5.6.X

haorlin:
This is the same fix as [https://github.com/sksamuel/elastic4s/issues/948](url) being applied to version 5.6.X


haorlin:
@sksamuel Would it be possible to merge this in sometime this week? Looking to implement a feature that requires percentiles to work. Thank you!!

sksamuel:
Done.
 Thanks.

Issue #1692, opened 2019-04-15
elastic4s 6.5.x cannot pass PorterStemTokenFilter as an analyzer


MattRMcFarland:
While using the elastic4s library version 6.5.1 for scala 2.11 against elasticsearch version 6.5.0, 6.5.4 and 7.0.0, I cannot apply the PorterStemTokenFilter to my customer analyzer.

When I attempt to do so, I receive the following error message
```
{
    "error": {
        "reason": "Custom Analyzer [nameAnalyzer] failed to find filter under name [porterStem]",
        "root_cause": [
            {
                "reason": "Custom Analyzer [nameAnalyzer] failed to find filter under name [porterStem]",
                "type": "illegal_argument_exception"
            }
        ],
        "type": "illegal_argument_exception"
    },
    "status": 400
}
```

I believe the root cause of the issue is that the elastic4s's mapping to the filter is `"porterStem"` which elasticsearch does not recognize. If you pass `"porter_stem"`, the analyzer is created as expected.

A fix would be to change the definition of `name` for PorterStemTokenFilter to `porter_stem` [here](https://github.com/sksamuel/elastic4s/blob/release/6.5.x/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/analyzers/TokenFilter.scala#L45) as per the elastic search documentation [here](https://www.elastic.co/guide/en/elasticsearch/reference/6.5/analysis-porterstem-tokenfilter.html).

MattRMcFarland:
Pull created [here](https://github.com/sksamuel/elastic4s/pull/1693)

PR #1693, opened 2019-04-15
rename PorterStemTokenFilter's name to "porter_stem"

MattRMcFarland:
`"porterStem"` is not to spec for Elasticsearch 6.5.x and above as I outlined in the issue reported [here](https://github.com/sksamuel/elastic4s/issues/1692).

`"porter_stem"` is the correct value defined by elasticsearch.


sksamuel:
Can we get this against master too?

MattRMcFarland:
@sksamuel sure! https://github.com/sksamuel/elastic4s/pull/1700

Issue #1697, opened 2019-04-19
Wrong request endpoint used in CatAllocationHandler


WangTaoTheTonic:
```
  implicit object CatAllocationHandler extends Handler[CatAllocation, Seq[CatAllocationResponse]] {

    override def build(request: CatAllocation): ElasticRequest =
      ElasticRequest("GET", "/_cat/aliases", Map("v" -> "", "format" -> "json", "bytes" -> "b"))
  }
```

As codes says, it used `/_cat/aliases` in CatAllocationHandler, which is wrong

PR #1698, opened 2019-04-19
use right endpoint in CatAllocationHandler 

WangTaoTheTonic:
https://github.com/sksamuel/elastic4s/issues/1697


WangTaoTheTonic:
Seems like sth wrong with build env.

sksamuel:
:+1:

PR #1700, opened 2019-04-22
rename PorterStemTokenFilter's name to "porter_stem"

MattRMcFarland:
[This is a copy of this change against 6.5.x]

`"porterStem"` is not to spec for Elasticsearch 6.5.x and above as I outlined in the issue reported [here](https://github.com/sksamuel/elastic4s/issues/1692).

`"porter_stem"` is the correct value defined by elasticsearch.


sksamuel:
:+1:

PR #1701, opened 2019-04-23
Change: Update Monix Version to 3.x.x

Yogeshberia:


sksamuel:
Thanks for the PR but I'd rather not bump to a pre-release version.

PR #1702, opened 2019-04-23
add missing weight attribute on random score function

larochef:
DSL allows to write things like 

```scala
randomScore(seed = 123456).fieldName("myField").weight(42D)
```

but in the generated json, the weight doesn't appear, as it does for other scoring functions.


sksamuel:
:+1:

PR #1707, opened 2019-04-24
Remove types

kimxogus:
Removed types for `search`, `indexInto`, `updateById` and many others except some api supports `include_type_name`.
There are still many test failures and need to be fiexed.


sksamuel:
You've done a lot of work here - and it's appreciated, but types in APIs are just deprecated in 7, not removed, right? So if we merge this, there's no way to query by type in indexes that already support them?

sksamuel:
After doing some upgrade work myself, I've decided it's best to just ditch types in many places, so I've merged the PR and I can finish it off.

kimxogus:
Thanks :)
It seems some apis with document type are still working but some don't and most of them are removed in official document.
I tried to maintain them and just add `deprecated` annotation at first, but it can also lead users to keep old style index structure and may cause many type related problems.

sksamuel:
It's certainly a confusing situation ! :)

PR #1726, opened 2019-05-13
(akka client) do not read response if method is HEAD

### LINKED ISSUE #1704: timeout with akka-http client doesn't on aliasExists
I'm getting timeouts when doing some `aliasExists` requests with the akka-http client.

After looking at the code, my feeling is that it's an issue with the HEAD requests that will not give any content in the response entity, and so, the source merging, to retrieve the response just waits indefinitely, since there is nothing to read.

what do you think about it?

larochef:
When doing head requests, ES sends the Content-Length header.
In order to avoid the client trying to read a response that will never
be sent, ignore response for HEAD operations

fixes #1704


sksamuel:
Awesome.

PR #1728, opened 2019-05-16
Fix `FetchSourceContext` case class

avdv:
Using `Array` properties in case classes destroys the ordinarily expected
equality and immutablity properties of case classes.

This change makes `FetchSourceContext` instances comparable and immutable.


sksamuel:
Hmm travis failed. Can you take a look?

avdv:
> Hmm travis failed. Can you take a look?

Sure, I fixed it up.

Note, I tried to make this change binary and (mostly) source compatible. It is binary compatible, but not source compatible in one respect:
```scala
val exc = Array("prop")

FetchSourceContext(false, excludes = exc) // no default for `includes` when using arrays
```
I guess it's not such a big deal though.

sksamuel:
Great, thanks!

PR #1736, opened 2019-05-24
add support in ElasticError for failed shards

philvarner:
Elastic sometimes reports error detail as part of the "failed_shards" error message attribute.  This PR adds support for parsing that part of the hierarchy, so users can programmatically access that part of the error message.  For example, the test case shows the error message from elastic when you do a geo_shape query with a Polygon that has a self-intersecting edge or multiple duplicated points.


sksamuel:
Awesome thanks.

PR #1747, opened 2019-06-05
handle Product in XContentBuilder, use XContentBuilder.autofield in SourceAsContentBuilder

dontgitit:
`SourceAsContentBuilder` didn't work for `BigDecimal`. It seems that `SourceAsContentBuilder.apply` was basically a copy of `XContentBuilder`, so switch to calling `autofield` instead and be done with it.

Only issue is, `XContentBuilder` didn't handle `Product`s before. I'm not sure if handling that can cause any negative side effects.


dontgitit:
Actually, on further inspection, looks like the `SearchHit` returned actually contains a `java.math.BigInteger`, which I don't see handled anywhere in this library yet...

sksamuel:
:+1: 

Maybe we can add BigInt support too.

PR #1748, opened 2019-06-06
Added support for _seq_no and _primary_term fields

sgrzeska:
Due to the fact that internal versioning was replaced by a new method based on sequence numbers I added support for it.


sksamuel:
:+1:

Issue #1710, opened 2019-04-26
Add `retryFailure` function to `SubscriberConfig` Reactive Streams subscriber


reeselevine:
When some requests in a bulk request fail, it would be useful to be able to specify if the failures should be retried. So I suggest adding another function to the `SubscriberConfig` which takes in a `BulkResponseItem` and returns `true` if the failure should be retried and `false` if not. It could default to always returning `true`, so would be backwards compatible with the current implementation.

PR #1755, opened 2019-06-10
Enhance handling of failures with retries in BulkIndexingSubscriber

reeselevine:
### This PR attempts to accomplish two things:
- Fix subscription requesting when some items in a bulk request fail
- Add a config function `retryFailure` that decides whether a bulk request item failure should be retried (https://github.com/sksamuel/elastic4s/issues/1710)

### Subscription Requesting
In the current implementation, if some items in a bulk request fail, the `BulkIndexingSubscriber` will end up requesting more items twice, once when handling the items that succeeded and again when handling the items that failed. This PR fixes that by only requesting more items if all items in a bulk request succeed, or if some items fail and there are no more attempts left.

### Retry Failures
Sometimes, an item in a bulk request will fail for reason that a client may not want to retry (for example, a version conflict error when using external versioning). This PR adds a function that takes in a `BulkResponseItem` and the original item and returns `true` if the item should be retried. This is backwards compatible by setting a default of always retrying.

I know that you've started implementing an Akka streams Sink @sksamuel which might replace this implementation, but until that's ready I think these additions would be helpful.

### DIFF
--- a/elastic4s-http-streams/src/main/scala/com/sksamuel/elastic4s/streams/BulkIndexingSubscriber.scala
+++ b/elastic4s-http-streams/src/main/scala/com/sksamuel/elastic4s/streams/BulkIndexingSubscriber.scala
@@ -68,9 +68,9 @@ object BulkActor {
 
   case object ForceIndexing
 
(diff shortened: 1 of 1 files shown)


sksamuel:
Awesome.

PR #1764, opened 2019-06-25
added ignore unavailable indices option to msearch

alepro:
Sometimes we need to make msearch and ignore unavailable indices here is possible solution.


sksamuel:
:+1:

PR #1786, opened 2019-07-11
Customization of HttpRequests in AkkaHttpClient

sgrzeska:
The purpose of this pull request is to add support for customization of HttpRequests in AkkaHttpClient by adding requestCallback to AkkaHttpClientSettings. It can then be used for example to add basic authentication header to requests:
`  val addBasicAuthorization: HttpRequest => HttpRequest = { request =>
    esConfig.credentials match {
      case None => request
      case Some(bc) =>
        val authorization = headers.Authorization(BasicHttpCredentials(bc.username, bc.password))
        request.copy(headers = request.headers :+ authorization)
    }
  }

  val settings = AkkaHttpClientSettings(
    https                = esConfig.uri.options.getOrElse("ssl", "false") == "true",
    hosts                = esConfig.uri.hosts.map { case (host, port) => s"$host:$port" }.toVector,
    queueSize            = 1000,
    poolSettings         = poolSettings,
    blacklistMinDuration = esConfig.blacklistMinDuration,
    blacklistMaxDuration = esConfig.blacklistMaxDuration,
    maxRetryTimeout      = esConfig.maxRetryTimeout,
    requestCallback = addBasicAuthorization
  )`

### DIFF
--- a/elastic4s-client-akka/src/main/scala/com/sksamuel/elastic4s/akka/AkkaHttpClient.scala
+++ b/elastic4s-client-akka/src/main/scala/com/sksamuel/elastic4s/akka/AkkaHttpClient.scala
@@ -218,7 +218,7 @@ class AkkaHttpClient private[akka] (
 
   private def toRequest(request: ElasticRequest,
                         host: String): Try[HttpRequest] = Try {
-    HttpRequest(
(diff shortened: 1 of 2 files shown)


sksamuel:
thank you!

PR #1788, opened 2019-07-12
Optimize memory usage in index request builder (7.1.x)

maxcom:
We observe OutOfMemoryError in our indexing application. This fix optimizes memory usage for indexing request builder. It fixes our problem and lowers memory and CPU usage for similar applications.


sksamuel:
7.1.2 released.

PR #1792, opened 2019-07-15
Allowing field mappings

fidesachates:
@sksamuel 

Couldn't find this feature: https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-get-field-mapping.html

This is PR adds the above functionality; it also removes `local` method from `GetMappingRequest` as I couldn't find any usage of it

### DIFF
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/indexes/MappingHandlers.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/indexes/MappingHandlers.scala
@@ -15,7 +15,7 @@ trait MappingHandlers {
           val raw = ResponseHandler.fromResponse[Map[String, Map[String, Map[String, Any]]]](response)
           val raw2 = raw.map {
             case (index, types) =>
-              val mappings = types("mappings").getOrElse("properties", Map.empty)
+              val mappings = types("mappings").getOrElse("properties", types("mappings"))
               IndexMappings(index, mappings.asInstanceOf[Map[String, Any]])
           }.toSeq
           Right(raw2)
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/indexes/MappingHandlers.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/indexes/MappingHandlers.scala
@@ -15,7 +15,10 @@ trait MappingHandlers {
           val raw = ResponseHandler.fromResponse[Map[String, Map[String, Map[String, Any]]]](response)
           val raw2 = raw.map {
             case (index, types) =>
-              val mappings = types("mappings").getOrElse("properties", types("mappings"))
+              val mappings = types("mappings") match {
+                case Map.empty => Map.empty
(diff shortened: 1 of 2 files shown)


elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/indexes/MappingHandlers.scala
```diff
@@ -15,7 +15,7 @@ trait MappingHandlers {
           val raw = ResponseHandler.fromResponse[Map[String, Map[String, Map[String, Any]]]](response)
           val raw2 = raw.map {
             case (index, types) =>
-              val mappings = types("mappings").getOrElse("properties", Map.empty)
+              val mappings = types("mappings").getOrElse("properties", types("mappings"))
```
fidesachates: `GetAllMappings` has an output with "properties"
```
{
    "index": {
        "mappings": {
            "properties": {
                "first_attribute": {
```

But when you look by fields, it doesn't have the properties field. It looks like:

```
{
    "index": {
        "mappings": {
            "first_attribute": {
```

I tested on 7.1.1 

To be frank, I'm not sure in what circumstance we'd expect properties to not be there in `GetAllMappings` so I removed the empty map.

elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/indexes/MappingHandlers.scala
```diff
@@ -15,7 +15,10 @@ trait MappingHandlers {
           val raw = ResponseHandler.fromResponse[Map[String, Map[String, Map[String, Any]]]](response)
           val raw2 = raw.map {
             case (index, types) =>
-              val mappings = types("mappings").getOrElse("properties", Map.empty)
+              val mappings = types("mappings") match {
+                case Map.empty => Map.empty
+                case nonEmptyMap: Map[string, Any] => nonEmptyMap.getOrElse("properties", nonEmptyMap)
+              }
```
fidesachates: `GetAllMappings` has an output with "properties"
```
{
    "index": {
        "mappings": {
            "properties": {
                "first_attribute": {
```

But when you look by fields, it doesn't have the properties field. It looks like:

```
{
    "index": {
        "mappings": {
            "first_attribute": {
```

I tested on 7.1.1

elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/indexes/MappingHandlers.scala
```diff
@@ -15,7 +15,10 @@ trait MappingHandlers {
           val raw = ResponseHandler.fromResponse[Map[String, Map[String, Map[String, Any]]]](response)
           val raw2 = raw.map {
             case (index, types) =>
-              val mappings = types("mappings").getOrElse("properties", Map.empty)
+              val mappings = types("mappings") match {
+                case Map.empty => Map.empty
```
fidesachates: In 7.1.1, if there are no mappings, the map is empty so rather than using the existence of "properties", just test to see if it's an empty map.

fidesachates:
Need to figure out edge cases before I reopen this pr

PR #1772, opened 2019-06-29
Scala 2.13.0 WIP

rkrzewski:
Fixes #1668 

Changes are source compatible with Scala 2.12 and 2.11. All tests are passing.

I had to sprinkle `toMap` and `toSeq` in multiple because of 2.13 collections migration. Most notably `Map.mapValues` returns a `MapView` now. In places where equality is checked, the view needs to be forced manually.

2.13 introduced changes to varargs treatment. I had to introduce `doc(field: (String, Any))` method taking a single tuple, because I got ambiguities errors between `doc(field: (String, Any)*)` and `def doc[T: Indexable](t: T)`.

Pending issues
- monix submodule is temporarily disabled because no build for Scala 2.13 is available yet
- cats, cats-effect and circe dependencies are at milestone releases

PR #1793, opened 2019-07-15
[6.5.x] inital support for scala 2.13 with a lot of modules disabled

jtjeferreira:
reused changes from https://github.com/sksamuel/elastic4s/pull/1772

pending issues:
* re-enable modules, but disable cross compilation for those


solicode:
Now that Cats 2.0.0 is out with 2.13 support, I think completing this might be possible now.

PR #1797, opened 2019-07-16
Make N-gram tokenizer work in 7.x

peri4n:
Why:

* Some tokenizer where wrapped in an object named after the tokenizer
  name. When serializing all used tokenizers within an analysis the name
  of the tokenizer gets written twice. They generated JSON gets rejected
  by Elastic Search.

How:

* Curiously, some tokenizer got serialized correctly (that is why the
  tests were green), so the only thing that was done to remove the
  object wrapping from the tokenizers that had it.


peri4n:
Hey @sksamuel, can you have a look at this and point me to things you want changed, if there are any?

sksamuel:
thank you!

PR #1802, opened 2019-07-19
Allowing field mappings

fidesachates:
@sksamuel 

Couldn't find this feature: https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-get-field-mapping.html

Originally I had opened a pr that merges this functionality into the existing `GetMappingRequest` and modified the respective `MappingHandler`. I realized that at the end of the day, I judged the two requests and responses to be just different enough to warrant a separate `Request` and thus a separate `Handler`. 

This PR also removes `local` method from `GetMappingRequest` as I couldn't find any usage of it

### DIFF
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/indexes/MappingHandlers.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/indexes/MappingHandlers.scala
@@ -1,15 +1,19 @@
 package com.sksamuel.elastic4s.requests.indexes
 
-import com.sksamuel.elastic4s.requests.mappings.{GetMappingRequest, PutMappingRequest}
+import com.sksamuel.elastic4s.requests.mappings.{GetFieldMappingRequest, GetMappingRequest, PutMappingRequest}
 import com.sksamuel.elastic4s._
 
 case class IndexMappings(index: String, mappings: Map[String, Any])
+
+case class IndexFieldMapping(index: String, fieldMappings: Seq[FieldMapping])
+case class FieldMapping(fullName:String, mappings: Map[String, Any])
 
 trait MappingHandlers {
 
   implicit object GetMappingHandler extends Handler[GetMappingRequest, Seq[IndexMappings]] {
 
     override def responseHandler: ResponseHandler[Seq[IndexMappings]] = new ResponseHandler[Seq[IndexMappings]] {
+
       override def handle(response: HttpResponse) : Either[ElasticError, Seq[IndexMappings]] = response.statusCode match {
         case 201 | 200       =>
           val raw = ResponseHandler.fromResponse[Map[String, Map[String, Map[String, Any]]]](response)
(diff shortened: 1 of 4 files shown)


elastic4s-tests/src/test/scala/com/sksamuel/elastic4s/requests/mappings/MappingHttpTest.scala
```diff
@@ -74,5 +84,55 @@ class MappingHttpTest extends WordSpec with DockerTests with Matchers with Befor
...
+        getMapping("indexnopropsempty")
+      }.await.result
+
+      val properties = mappings.find(_.index == "indexnopropsempty").get.mappings
+
+      properties shouldBe Map.empty
+    }
+  }
+
+  "field mapping get" should {
```
fidesachates: Test cases for when we ask for a specific field in a mapping get request. These are a mirror of the original two tests in addition to the third one I added for when we get all mappings.

elastic4s-tests/src/test/scala/com/sksamuel/elastic4s/requests/mappings/MappingHttpTest.scala
```diff
@@ -74,5 +84,55 @@ class MappingHttpTest extends WordSpec with DockerTests with Matchers with Befor
 
       properties shouldBe Map.empty
     }
+
+    "handle properly completely empty mapping" in {
```
fidesachates: New edge case test when the mapping is completely empty

elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/indexes/MappingHandlers.scala
```diff
@@ -28,12 +32,49 @@ trait MappingHandlers {
...
+  implicit object GetFieldMappingRequest extends Handler[GetFieldMappingRequest, Seq[IndexFieldMapping]] {
+    override def responseHandler: ResponseHandler[Seq[IndexFieldMapping]] = new ResponseHandler[Seq[IndexFieldMapping]] {
+
+      override def handle(response: HttpResponse) : Either[ElasticError, Seq[IndexFieldMapping]] = response.statusCode match {
+        case 201 | 200       =>
+          val raw = ResponseHandler.fromResponse[Map[String, Map[String, Map[String, Map[String, Any]]]]](response)
+          Right(raw.map {
+            case (index, types) =>
+              val mappings = types("mappings").map {
+                case (_, mapping) => FieldMapping(mapping("full_name").toString, mapping("mapping").asInstanceOf[Map[String, Any]])
```
fidesachates: This method is largely a copy pasta, but this part is different as it's deserializing an object that looks like this:

```
{
    "index": {
        "mappings": {
            "a": {
                "full_name": "a",
                "mapping": {
                    "a": {
                        "type": "text",
                        "store": true,
                        "analyzer": "whitespace"
                    }
                }
            }
        }
    }
}
```

elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/mappings/GetMappingRequest.scala
```diff
@@ -1,8 +1,5 @@
 package com.sksamuel.elastic4s.requests.mappings
 
 import com.sksamuel.elastic4s.Indexes
-import com.sksamuel.exts.OptionImplicits._
 
-case class GetMappingRequest(indexes: Indexes, local: Option[Boolean] = None) {
-  def local(local: Boolean): GetMappingRequest = copy(local = local.some)
```
fidesachates: As I said, removing because I couldn't find a usage of it.

elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/indexes/MappingHandlers.scala
```diff
@@ -28,12 +32,49 @@ trait MappingHandlers {
...
+    }
+
+    override def build(request: GetFieldMappingRequest): ElasticRequest = {
+      val baseEndpoint = request.indexes match {
+        case Indexes(Nil)       => "/_mapping"
+        case Indexes(indexes)   => s"/${indexes.mkString(",")}/_mapping"
+      }
+
+      val endpoint = request.fields match {
+        case Nil => throw new IllegalArgumentException("Empty fields in GetFieldMappingRequest. Use GetMappingRequest instead.")
```
fidesachates: Perhaps this isn't the right thing to do? A field mapping request requires at least one field;.

sksamuel:
Sorry for the late merge, but great work.

PR #1808, opened 2019-07-29
value type of after of composite aggregation should be Any

EugeneChung:
- As the official example says, the value can be number.
"after": { "date": 1494288000000, "product": "mad max" }
- https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-composite-aggregation.html#_after


sksamuel:
thank you!

Issue #1678, opened 2019-03-28
Add possibility to create a client with sniffing


yannick-cw:
Hi :wave: 
When I want to use cluster sniffing with the client now, I have to create the sniffing through the java api
```
    val restClient: RestClient = RestClient
      .builder(new HttpHost"host", "port", "http"))
      .build()
    Sniffer.builder(restClient).build()
    val client = ElasticClient.fromRestClient(restClient)
```
with adding the `elasticsearch-rest-client-sniffer` dependency.

Do you think it could be possible to integrate the sniffing into elastic4s or into a elastic4s.sniffing module? If yes I would be happy to help out.

yannick-cw:
Actually I just run into a prod bug, as I was using the Sniffing client wrongly :/
The java mutable api was confusing me. So I'll create a small scala package now with a sniffing client. If there is place for it in elastic4s please let me know :)

sksamuel:
Please do make a PR.

PR #1830, opened 2019-08-13
Add submodule for a sniffing client

yannick-cw:
As proposed in #1678 this adds a client with sniffing support.

I am not sure about testing this, as it would need some integration tests with a running elasticsearch in docker I guess.


sksamuel:
:+1:

PR #1831, opened 2019-08-14
Add search as you type field & bool_prefix query type

jackpf:
In version 7.3.0 it seems we have a new `search_as_you_type` field type, and a new `match_bool_query` query type:

https://www.elastic.co/guide/en/elasticsearch/reference/current/search-as-you-type.html
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-match-bool-prefix-query.html

We wanted to test this out, so I've added these types for a local build.

Here's a PR if you're interested :)


sksamuel:
Awesome thank you.

PR #1838, opened 2019-08-27
Add multiplexer token filter support

alexshadow007:
https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-multiplexer-tokenfilter.html


sksamuel:
:+1:

PR #1854, opened 2019-09-12
Add compound word token filter types into analysis package

nelusnegur:
Currently it's not possible to create an instance of `analysis.CompoundWordTokenFilter` because there are no defined values for the `analysis.CompoundWordTokenFilterType`. 

This PR aims to fix this issue by defining the types of a compounded word token filter supported by Elasticsearch, in the new `analysis` package.


nelusnegur:
The unrelated failing test was fixed, and I rebased onto master; so this PR is now ready to be reviewed. @sksamuel when you have time, please take a look 🙂

sksamuel:
Good work thanks.

PR #1863, opened 2019-09-14
SttpRequestHttpClient Interface Improvements

alexklibisz:
I tried using the sttp client and ran into a couple issues which I fixed in this PR. Specific changes are:

1. The current `SttpRequestHttpClient` class is constructed using the deprecated `ElasticsearchClientUri`.. if you are using fatal warnings in your project this means you can't use the `SttpRequestHttpClient` at all. So I switched to using the `ElasticNodeEndpoint` in the client constructor.
2. The current client class uses the global execution context internally by default. In my experience it's common to provide an execution context because you may want to tune it for specific performance characteristics. So I provided the execution context as an implicit parameter to the client class.
3. The current client class hard-codes the sttp backend internally in the class. There are several types of sttp backends and the user might want to share one across multiple client wrappers. So I also abstracted the sttp client out as an implicit parameter to the client class.
4. I added a default `apply` method in the client companion object which uses the global EC and the `AsyncHttpClientFutureBackend` by default to make it easy to use the client without thinking about those parameters.
5. For backwards compatibility, I added an alternative constructor that takes just the `ElasticsearchClientUri` and marked it as deprecated. This means people shouldn't have to change their code, but they will get warnings. I also hard-coded the "http" protocol similar to how it was originally implemented in the client.
6. I changed the method string pattern match for "POST" so that it calls `sttp.post` instead of `sttp.get`... maybe this was intentional but it seemed like a bug to me.

Please let me know your feedback.. I'm happy to iterate on the PR.


sksamuel:
I can't see the deprecated compatibility method.

alexklibisz:
> I can't see the deprecated compatibility method.

I forgot to push that commit, sorry, should be pushed now!

sksamuel:
Good work. I agreed with all the points you made in the PR.

alexklibisz:
@sksamuel Thanks! What's the release cadence for this project? Do you know approx. when this change would be up on maven?

sksamuel:
There's a problem with the 2.12 build I need to fix first.

PR #1865, opened 2019-09-17
Bump cats-effect to latest version (ie 2.0.0)

eric-taix-teads:


eric-taix-teads:
CI failed, but it seems that my PR is not the root cause of failed tests ? Any idea ?

sksamuel:
:+1:

PR #1867, opened 2019-09-18
fix TermsAggregation.excludeExactValues

jdcohen220:
The overloaded method that accepts a Set[String] argument was incorrectly including that set instead of excluding.


sksamuel:
:+1:

PR #1881, opened 2019-09-23
Remove unnecessary parsing in IndexContentBuilder

martmarch:
I noticed that the request source field in IndexContentBuilder is parsed from String to XContentBuilder, and then in BulkBuilderFn from XContentBuilder to String. My suggestion is to remove unnecessary parsing to XContentBuilder and return String in IndexContentBuilder. 

I'm not sure, whether it was a overlooking or it was made on purpose.


sksamuel:
👍

PR #1882, opened 2019-09-26
QueryString query time_zone support

stogo:
Hi! Could you please review these changes?

I added missing time_zone parameter to QueryStringQuery

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#query-string-top-level-params
> time_zone
> (Optional, string) Coordinated Universal Time (UTC) offset or IANA time zone used to convert date values in the query string to UTC.
> 
> Valid values are ISO 8601 UTC offsets, such as +01:00 or -08:00, and IANA time zone IDs, such as America/Los_Angeles.


sksamuel:
👍

EliseevAndrew:
Could you add this parameter to 6.3.x release?

PR #1896, opened 2019-10-07
add support for new analysis package

martmarch:
Hi,

I've added support for new analysis package in CreateIndexTemplateRequest, analogously to what was done in CreateIndexRequest, since all the analysis functions in that object had deprecated annotations.


sksamuel:
Great work thank you.

PR #1906, opened 2019-10-17
Remove type field from PreIndexedShapes since it is deprecated

jdcohen220:
The `type` field on preindexed geo_shape queries is deprecated, resulting in warnings from ES. removed the unnecessary type parameter from the PreIndexedShape entity.


sksamuel:
:+1:

PR #1917, opened 2019-10-24
Close the ElasticSource output port when no more elements is scroll response

Fraer:
Small fix: the ElasticSource graph stage never closes the output port when there is no more elements in the scroll response, so the stream never ends.


sksamuel:
:+1:

PR #1925, opened 2019-11-01
Replace SearchError with ElasticError to be able to read the root cause

Dbzman:
Hey,

it happened sometimes that we misconfigured our searches (for instance we configured an aggregation with a non-fitting field name). The problem was that we only got a generic `all_shards_failed` error with no useful information. The root cause has not been shown. I saw that `ElasticError` has been introduced recently which improves the situation but it wasn't used in the `MultiSearchHandler`. I removed `SearchError` in the `MultiSearchHandler` and replaced it with `ElasticError`.

Cheers!
Timo


sksamuel:
:+1:

PR #1930, opened 2019-11-05
Fix issues #1922 and #1799

### LINKED ISSUE #1799: Scroll is not closed when the first response is empty
In [PublishActor](../blob/c0fdfe696898d3fcc4ce0250fc04a6ea723f45f1/elastic4s-http-streams/src/main/scala/com/sksamuel/elastic4s/streams/ScrollPublisher.scala#L161) `clearScroll` uses the old `scrollId` value rather then new from `resp.result.scrollId`, which is required by elastic documentation. This becomes especially important when the first response is empty, since `scrollId` is not yet initialized and the scroll will not be closed.

### LINKED ISSUE #1922: The PublishActor should try to clearScroll on postStop
Currently, the `PublisherActor` only clears the scroll id when there's an empty message (https://github.com/sksamuel/elastic4s/blob/master/elastic4s-http-streams/src/main/scala/com/sksamuel/elastic4s/streams/ScrollPublisher.scala#L161)

Note that there are cases where the actor dies without clearing the `scrollId`, which can lead to a ton of open contexts (e.g. an overloaded node can start returning timeouts, which would kill the actor without removing the scrolling context).

I think that the actor should always try to call the `clearScroll` on `onStop`, as a best effort.

duartepinto:
* Update `scrollId` even if the response `isEmpty`
* Override `postStop` to try to `clearScroll` before stopping actor 

Should close #1922. Should close #1799

### DIFF
--- a/elastic4s-http-streams/src/main/scala/com/sksamuel/elastic4s/streams/ScrollPublisher.scala
+++ b/elastic4s-http-streams/src/main/scala/com/sksamuel/elastic4s/streams/ScrollPublisher.scala
@@ -159,7 +159,6 @@ class PublishActor(client: ElasticClient, query: SearchRequest, s: Subscriber[_ 
       logger.debug("Response from ES came back empty; this means no more items upstream so will complete subscription")
       scrollId = resp.result.scrollId.getOrElse(scrollId)
       s.onComplete()
-      client.execute(clearScroll(scrollId))
       logger.debug("Stopping publisher actor")
       context.stop(self)
     // more results and we can unleash the beast (stashed requests) and switch back to ready mode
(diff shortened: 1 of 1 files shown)


elastic4s-http-streams/src/main/scala/com/sksamuel/elastic4s/streams/ScrollPublisher.scala
```diff
@@ -157,6 +157,7 @@ class PublishActor(client: ElasticClient, query: SearchRequest, s: Subscriber[_
     // if we had no results from ES then we have nothing left to publish and our work here is done
     case Success(resp: RequestSuccess[SearchResponse]) if resp.result.isEmpty =>
       logger.debug("Response from ES came back empty; this means no more items upstream so will complete subscription")
+      scrollId = resp.result.scrollId.getOrElse(scrollId)
       s.onComplete()
       client.execute(clearScroll(scrollId))
```
fggarcia: I believe that you don't this line.

duartepinto: Removed in 71c52a7

sksamuel:
awesome thanks

PR #1938, opened 2019-11-14
use trusty image in travis build

kimxogus:
https://travis-ci.community/t/error-installing-oraclejdk8-expected-feature-release-number-in-range-of-9-to-14-but-got-8/3766/6

Also bumped scala patch versions


sksamuel:
👍

PR #1944, opened 2019-11-22
Add missing_bucket to all CompositeAggregation value sources

Tomassino:
I've added the "missing_bucket" property to all value sources of a CompositeAggregation. The current version of elastic4s only had it for `DateHistogramValueSource`


sksamuel:
:+1:

Issue #1926, opened 2019-11-03
Creating and reading ingest processors?


alexklibisz:
Hi, I'm working on a project that leverages [ES ingest processors](https://www.elastic.co/guide/en/elasticsearch/reference/current/ingest-processors.html).

I looked through the various request types in this lib and couldn't find anything for making and retrieving processors. I'm wondering if maybe I'm overlooking them somewhere?

If not, would it be valuable to contribute them to the library? I have a working `Handler` in my code that I could pull out and PR into elastic4s. LMK if that would be helpful, and if so, which package they should go in.

sksamuel:
A PR would be fantastic, thank you.

ckcedrickeller:
Hi,
also looking for pipelines.
I would like to use existing processors, but I would need to be able to use pipelines : [ES Date Index Name Processor](https://www.elastic.co/guide/en/elasticsearch/reference/current/date-index-name-processor.html).

Is it possible to use pipelines with the actual version of elastic4s ? didn't find how.
Any pointer would be very much appreciated.

alexklibisz:
@ckcedrickeller I haven't had the time to integrate it in the elastic4s codebase, but this is how I'm doing it in my project:

```
  case class Processor(name: String, opts: ProcessorOptions)
  case class Pipeline(description: String, processors: Seq[Processor])
  case class PipelineRequest(name: String, pipeline: Pipeline)
  case class PipelineResponse(acknowledged: Boolean)

  object PipelineRequest {
    implicit object PipelineRequestHandler extends Handler[PipelineRequest, PipelineResponse] {
      override def build(t: PipelineRequest): ElasticRequest = {
        // See: https://www.elastic.co/guide/en/elasticsearch/reference/current/ingest-processors.html
        val body: Json = Json.fromJsonObject(
          JsonObject(
            "description" -> Json.fromString(t.pipeline.description),
            "processors" -> Json.fromValues(t.pipeline.processors.map { p =>
              Json.fromJsonObject(
                JsonObject(p.name -> JsonFormat.toJson(p.opts))
              )
            })
          ))
        val endpoint = s"_ingest/pipeline/${t.name}"
        ElasticRequest("PUT", endpoint, HttpEntity(body.noSpaces))
      }
    }
  }
```

[...]

PR #1946, opened 2019-11-26
Ingest requests and handlers

alexklibisz:
This addresses: https://github.com/sksamuel/elastic4s/issues/1926

Some specific places where feedback would be great:
- Not sure the request names are great. Some of the other requests include the http verb (e.g. `PutMappingRequest`), while others do not (e.g. `IndexRequest`). I went with the former but can easily change.
- Does the package make sense? I figured `requests.ingest` makes sense since the endpoints are all prefixed with `_ingest`.
- Passing the `configuration` json as a raw string doesn't seem great, but there's precedent, like the `source` field in the `IndexRequest` case class.

For the tests I just made sure that the scala-based requests match the pure http/json requests in the docs for a couple common pipelines/processors. I can make this more rigorous if needed.

Also, thanks to @eliza-jane for helping out at the Boston scala project night!

### DIFF
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/ingest/Processor.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/ingest/Processor.scala
@@ -1,0 +1,3 @@
+package com.sksamuel.elastic4s.requests.ingest
+
+case class Processor(name: String, configuration: String)
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/ingest/Processor.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/ingest/Processor.scala
@@ -1,3 +1,41 @@
 package com.sksamuel.elastic4s.requests.ingest
 
-case class Processor(name: String, configuration: String)
+import com.sksamuel.elastic4s.{XContentBuilder, XContentFactory}
+
+/**
+  * Abstract representation of a processor with a constant name (e.g. "geoip") and an [[XContentBuilder]] that
+  * constructs the body (e.g. { "field" : "ip" } ).
+  */
+trait Processor {
+  def name: String
+  def buildProcessorBody(): XContentBuilder
+}
+
+/**
+  * Processor defined by its name and raw Json options.
+  */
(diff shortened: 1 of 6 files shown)


alexklibisz:
Looks like it failed on an unrelated error. 

image


Any tips @sksamuel ?

alexklibisz:
@sksamuel Let me know if there's anything I can do to help get this across the finish line.. Thanks for your work on this project!

sksamuel:
We could merge this as is.
But like you I don't think the string json for configuration is great.
Perhaps we add:

```scala
trait ProcessorConfiguration  {
  def config:String
}
```
Which returns json when config is invoked.

Then we can add strongly typed versions for the built in processors like ScriptProcessor, DateProcessor etc (you could add a couple to show it works, the rest could follow later).

alexklibisz:
> We could merge this as is.
> But like you I don't think the string json for configuration is great.
> Perhaps we add:
> 
> ```scala
> trait ProcessorConfiguration  {
>   def config:String
> }
> ```
> 
> Which returns json when config is invoked.
> 
> Then we can add strongly typed versions for the built in processors like ScriptProcessor, DateProcessor etc (you could add a couple to show it works, the rest could follow later).

Good idea. I'll take a pass at that tonight.

elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/ingest/Processor.scala
```diff
@@ -0,0 +1,41 @@
...
+
+import com.sksamuel.elastic4s.{XContentBuilder, XContentFactory}
+
+/**
+  * Abstract representation of a processor with a constant name (e.g. "geoip") and an [[XContentBuilder]] that
+  * constructs the body (e.g. { "field" : "ip" } ).
+  */
+trait Processor {
+  def name: String
+  def buildProcessorBody(): XContentBuilder
```
alexklibisz, line 11: @sksamuel I followed the same structure as the `CustomQuery#buildQueryBody` method: https://github.com/sksamuel/elastic4s/blob/919875f4280a96cbf9307bb61a3386437a0cf33f/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/searches/queries/Query.scala#L12

sksamuel:
Nice.

PR #1947, opened 2019-11-27
Fix endpoint path in ValidateHandler

tobiatesan:
Without the initial "/" validate would produce an invalid HTTP request
-- i.e. GET foo instead of GET /foo -- that would e.g. upset
proxies like Nginx.

See https://www.ietf.org/rfc/rfc2396.txt page 10 or
https://stackoverflow.com/questions/27638278/do-http-paths-have-to-start-with-a-slash


sksamuel:
Thanks!

PR #1960, opened 2019-12-13
[6.7.x] add rest_total_hits_as_int param

kimxogus:
In rolling upgrade to es 6.x to 7.x, `rest_total_hits_as_int` should be set `true` for elasticsearch-client@6 because `hits.total` changed to object in https://github.com/elastic/elasticsearch/pull/35849


sksamuel:
:+1:

PR #1961, opened 2019-12-14
6.2.x fix MultiSearchBuilder to properly set search_type and preference header fields

TF2274:
This fix is identical to the fix done in later branches.
My team cannot transition to later versions of this library because of the large amount of breaking changes in later versions of this library. At least not until we have the bandwidth.

That is my motivation for making a PR in an older branch to fix this bug. Until we have the bandwidth to go to a version of elastic4s (which is a tough decision given how often newer versions of this library change too much and require hours of work on our end) we have to hope that this bug fix can become a 6.2.x release or resort to hosting our own version of this library for the time being.

While this bug is patched in later versions of this library, it is not patched in earlier versions such as 6.2.x.
Elasticsearch took a "preference" parameter in multi-search queries back in 6.2 (and long before it), and the search types were snake-case, not camel case, long before 6.2. Therefore this is a valid bug fix.

### DIFF
--- a/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/search/MultiSearchBuilderFn.scala
+++ b/elastic4s-http/src/main/scala/com/sksamuel/elastic4s/http/search/MultiSearchBuilderFn.scala
@@ -1,7 +1,7 @@
 package com.sksamuel.elastic4s.http.search
 
 import com.sksamuel.elastic4s.json.XContentFactory
-import com.sksamuel.elastic4s.searches.MultiSearchDefinition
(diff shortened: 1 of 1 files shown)


sksamuel:
I've released 6.2.11 with this fix.
To be fair, a lot of the changes between versions are forced on this library by breaking changes in elasticsearch itself (eg, types -> no types).

TF2274:
Fair enough.
It's just a little bit frustrating.
I appreciate you accepting this PR and making a release so quickly. It'll let us knock out a bad defect without doing any 'janky' temporary work-around.

PR #1973, opened 2019-12-21
Add custom backwards compatible deserializer for Total

easel:
- Allows reading a number as produced by es6 vs the object produced by es7


easel:
This one is a little tough to test, unfortunately. The existing tests still pass, but getting the entire es7 compatible test suite to pass against es6 and automating it is pretty tricky. Ideas welcome.

easel:
Just as a point of order, I've compiled and deployed this to a couple of applications with both elastic6 and elastic7 and it nominally works.

sksamuel:
I like the approach. Thank you for the pr.

easel:
Thanks for the merge!

PR #1975, opened 2019-12-28
Fix routing parameter of bulk update and delete for ES 7.x and on

chadford:
Fixes #1955 

Tested for bulk deletes against ES 7.1

PR #1980, opened 2020-01-08
[7.0.x] Fix routing parameter of bulk update and delete

chadford:
Same fix as #1975


chadford:
Re-triggered the build a few times. Some jdk install issue while booting the virtual machine
```Expected feature release number in range of 9 to 15, but got: 8

The command "~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"" failed and exited with 3 during .```

chadford:
@sksamuel any ideas why this isn't passing on the branches? Doesn't seem related to the PR at all.

PR #1985, opened 2020-01-17
Fix Levenshtein and Jaro-Winkler string distance naming

Philippus:
Applying this PR will fix the naming of and the `valueOf`-method for the case objects of the Levenshtein and Jaro-Winkler string distance options. Right now they result in an `IllegalArgumentException("Illegal distance option ...")`.


sksamuel:
Nice thanks.

PR #1987, opened 2020-01-24
Add pinned query

Philippus:
Applying this PR will add the `pinned query` feature, which was introduced in Elasticsearch 7.4.0 (https://www.elastic.co/guide/en/elasticsearch/reference/7.4/query-dsl-pinned-query.html).


sksamuel:
Nice work.

PR #1994, opened 2020-02-06
[6.5.x] Add support for include_in_root|parent settings for nested fields

kirilyuro:
The `include_in_root` and `include_in_parent` settings for `nested` fields were removed from the elasticsearch documentation in version 2.0 (see: [1.7](https://www.elastic.co/guide/en/elasticsearch/reference/1.7/mapping-nested-type.html) vs [2.0](https://www.elastic.co/guide/en/elasticsearch/reference/2.0/nested.html)).
However these settings are still supported by all versions of elasticsearch (tested on current latest version 7.5.2), and a [discussion about removing support for these settings](https://github.com/elastic/elasticsearch/issues/12461) is stalled since May 2018.

Since this is still supported by elasticsearch, I think elastic4s should support this too.
Also, how are you managing porting to other releases? I think this should be ported to all other version as well. Should I open separate PRs for the other releases?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,4 +27,4 @@ script:
 - sbt clean test
 
 jdk:
-- oraclejdk8
+- openjdk8
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/mappings/CommonFieldBuilder.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/mappings/CommonFieldBuilder.scala
@@ -108,6 +108,8 @@ object FieldBuilderFn {
 
       case nested: NestedField =>
         nested.dynamic.foreach(builder.field("dynamic", _))
+        nested.includeInParent.foreach(builder.field("include_in_parent", _))
+        nested.includeInRoot.foreach(builder.field("include_in_root", _))
 
       case text: TextField =>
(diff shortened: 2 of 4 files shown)


kirilyuro:
@sksamuel 
Build is now passing after changing oraclejdk8 to openjdk8 in build env.
(Travis CI changed the default build env distribution from Trusty to Xenial, which does not support oraclejdk8 - https://travis-ci.community/t/install-of-oracle-jdk8-is-failing/4365/3).

sksamuel:
Thanks for the PR. Would appreciate the PR against the other branches too, that would be great!

kirilyuro:
@sksamuel 
The referenced PRs should cover porting to all other releases.

A note about pre-5.x releases:
These options are actually supported in the `2.1.x`, `2.3.x` and `2.4.x` releases.
Support was added in https://github.com/sksamuel/elastic4s/pull/488.
If someone will need them in pre-2.1.x releases, https://github.com/sksamuel/elastic4s/pull/488 should be back-ported.

kirilyuro:
@sksamuel 
Also, when can I expect a release of `6.5.x` with these changes?

sksamuel:
6.5.5 has been released to maven central.

Issue #1992, opened 2020-02-05
7.3.5: NoSuchElementException when creating an index with no "analyzer"


rogueai:
When upgrading from `7.3.1` to `7.3.5` we found that creating an index with settings, but no analyzer, results in a `NoSuchElementException` being thrown, and index creation fails.

Upon investigation it looks like the following is the root cause:
https://github.com/sksamuel/elastic4s/blob/f3341d9491eb4c64dc0db58b8e119cebdee4a809/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/indexes/IndexTemplateHandlers.scala#L85

Thanks!

tarangbhalodia:
I would like to work on this.

sksamuel:
Please do.

tarangbhalodia:
@sksamuel 
PR: https://github.com/sksamuel/elastic4s/pull/2004 is ready for review

PR #2004, opened 2020-02-20
fix NoSuchElementException when creating index

tarangbhalodia:
Fix for: https://github.com/sksamuel/elastic4s/issues/1992

### DIFF
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/indexes/IndexTemplateHandlers.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/indexes/IndexTemplateHandlers.scala
@@ -77,12 +77,12 @@ object CreateIndexTemplateBodyFn {
     create.order.foreach(builder.field("order", _))
     create.version.foreach(builder.field("version", _))
 
-    if (create.settings.nonEmpty || create.analysis.nonEmpty) {
+    if (create.settings.nonEmpty) {
       builder.startObject("settings")
       create.settings.foreach {
         case (key, value) => builder.autofield(key, value)
       }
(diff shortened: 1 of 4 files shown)


tarangbhalodia:
@sksamuel This PR is ready. Please let me know if any change required.

elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/indexes/IndexTemplateHandlers.scala
```diff
@@ -82,7 +82,7 @@ object CreateIndexTemplateBodyFn {
       create.settings.foreach {
         case (key, value) => builder.autofield(key, value)
       }
-      builder.rawField("analysis", AnalysisBuilder.build(create.analysis.get))
+      if(create.analysis.nonEmpty) builder.rawField("analysis", AnalysisBuilder.build(create.analysis.get))
```
Philippus: how about avoiding the whole `.get` with:
```
      create.analysis.foreach { analysis =>
          builder.rawField("analysis", AnalysisBuilder.build(analysis))
      }
```

sksamuel: Yep that would be better.

tarangbhalodia: `.foreach` sounds good. Let me change it.

tarangbhalodia: I've updated it.

sksamuel:
:+1:

Issue #2021, opened 2020-03-05
org.scalatestplus:mockito-1-10:3.1.1.0 cannot be resolved from standard repos


philvarner:
Running `sbt test` from master, I get:

[error] sbt.librarymanagement.ResolveException: Error downloading org.scalatestplus:mockito-1-10_2.12:3.1.1.0
[error]   Not found
[error]   Not found
[error]   not found: /Users/pvarner/.ivy2/local/org.scalatestplus/mockito-1-10_2.12/3.1.1.0/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/org/scalatestplus/mockito-1-10_2.12/3.1.1.0/mockito-1-10_2.12-3.1.1.0.pom
[error]   not found: /Users/me/.m2/repository/org/scalatestplus/mockito-1-10_2.12/3.1.1.0/mockito-1-10_2.12-3.1.1.0.pom
[error] 	at lmcoursier.CoursierDependencyResolution.unresolvedWarningOrThrow(CoursierDependencyResolution.scala:246)
[error] 	at lmcoursier.CoursierDependencyResolution.$anonfun$update$34(CoursierDependencyResolution.scala:215)

PR #2024, opened 2020-03-10
Fix CI build by using ScalatestPlusMockito artifact for 3.2 instead of 1.10

philvarner:
use ScalatestPlusMockito artifact for 3.2 instead of 1.10
remove ignored uses of MaxPermSize (causing warning msg)
update scala 2.12 and 2.13 patch versions  - this prevented me from compiling, throwing a "/packages" error

To resolve https://github.com/sksamuel/elastic4s/issues/2021


sksamuel:
Thanks Phil.

PR #2025, opened 2020-03-10
Remove unused and optimize imports

philvarner:
There are a large number of unused import warnings, this PR resolves all of them. This was done using IntelliJ's Optimize Imports, which also reorders imports. No specific imports were collapsed into `_`.

Draft PR for now, as this contains the build fixes in https://github.com/sksamuel/elastic4s/pull/2024

PR #2026, opened 2020-03-10
Upgrade es client to 7.6.1

philvarner:
Draft until #2024 and #2025 are merged. 

This upgrades the ES library to 7.6.1, updates the docker-compose.yml to the matching version, and fixes the one test that broke w/ the service upgrade b/c the error message is different as of 7.6.0.


philvarner:
Ready for PR pending CI.

PR #2049, opened 2020-04-02
Support dense_vector field type in mappings

alicefuzier:
Hi,

We need support for a [dense_vector](https://www.elastic.co/guide/en/elasticsearch/reference/current/dense-vector.html) field type in our mapping definition. This is a pretty basic attempt at providing that.
My knowledge of both elastic4s and elasticsearch isn't great, so apologies if this doesn't make completely sense. I'm happy to change if need be.

### DIFF
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/fields/fields.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/fields/fields.scala
@@ -367,4 +367,28 @@ case class IpField(name: String,
                    properties: Seq[ElasticField] = Nil,
                    store: Option[Boolean] = None) extends ElasticField
 
+case class DenseVectorField(name: String,
+                            dims: Int,
+                            analysis: Analysis = Analysis(),
+                            boost: Option[Double] = None,
+                            coerce: Option[Boolean] = None,
+                            copyTo: Seq[String] = Nil,
+                            docValues: Option[Boolean] = None,
+                            enabled: Option[Boolean] = None,
+                            fielddata: Option[Boolean] = None,
+                            fields: Seq[ElasticField] = Nil,
+                            format: Option[String] = None,
+                            includeInAll: Option[Boolean] = None,
+                            ignoreAbove: Option[Int] = None,
+                            ignoreMalformed: Option[Boolean] = None,
+                            index: Option[String] = None,
+                            indexOptions: Option[String] = None,
+                            norms: Option[Boolean] = None,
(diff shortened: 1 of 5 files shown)


sksamuel:
Thanks, I think the case class is good, but the best place for it is in the new ElasticField ADT.

alicefuzier:
I'm having a look at how to do this but I don't seem to understand how the `ElasticField` ADT is used at the moment? `TextField`, as an example, is in `package com.sksamuel.elastic4s.fields` as a subclass of `ElasticField` and in `package com.sksamuel.elastic4s.requests.mappings` as a subclass of `FieldDefinition`. Should I be doing the same and put it in both places? Also, is there a quick way of knowing which parameters in `DenseVectorField` make sense? The elastic documentation that I could  find doesn't seem to be very explanatory in this regard

sksamuel:
Unfortunately the elastic docs aren't always the best.

What you did by extending ElasticField is correct. Then you can use it by passing an instance to the properties field when creating a mapping.

alicefuzier:
I think what confuses me is how to use it with the DSL. We're using the `CreateIndexApi` like this 
```
elasticClient
      .execute {
        createIndex(index.name)
          .mapping { mapping }
...
```
where `mapping` is 
```
val mapping = properties(fields).dynamic(DynamicMapping.Strict)
```
and `fields` is a `Seq[FieldDefinition]`.
So we're calling the method `properties` on `CreateIndexApi` which as far as I can see wants `FieldDefinition` not `ElasticField`.
Also `CreateIndexRequest.mapping` wants a `MappingDefinition` which as far as I can see does not have a `properties: ElasticField` parameter (but it does have `fields: Seq[FieldDefinition]`).

Is the intention in the future to move everything over to `ElasticField`? Am I missing something very simple here?

alicefuzier:
Hi, just wondering if you have any comments on the above? This would be a useful feature for us (we are in fact using the forked version in production) and I would like to avoid this PR becoming stale...
Again apologies if my questions are a bit obvious. If you could point me to the relevant docs I would be happy to work it out myself.

sksamuel:
Sorry, I will get more details as soon as I can.

sksamuel:
So the MappingDefinition accepts the ElasticField types as well as the soon to be deprecated types. The `properties` DSL function, which just builds the mapping definition for you, needs to be overloaded to accept both types.

sksamuel:
I have merged this because I have realized I am talking about something not released yet that I am working on. So I apologize.

alicefuzier:
Ah thank you! I only just realised but this is great 😄

PR #2060, opened 2020-04-23
 scala 2.13 cross compile support for the 5.6.x branch

samzilverberg:
added support for scala 2.13 for the 5.6.x release.

did:

- started from release/5.6.x
- Build.scala 
  - added 2.13.1 to crossScalaVersions
  - update dependencies to ones that support 2.11 till 2.13
- compiled and fixed errors where needed
- ran tests on all cross scala versions

### DIFF
--- a/elastic4s-http-streams/src/main/scala/com/sksamuel/elastic4s/streams/ScrollPublisher.scala
+++ b/elastic4s-http-streams/src/main/scala/com/sksamuel/elastic4s/streams/ScrollPublisher.scala
@@ -150,7 +150,7 @@ class PublishActor(client: HttpClient,
     // more results and we can unleash the beast (stashed requests) and switch back to ready mode
     case Success(resp: SearchResponse) =>
       scrollId = resp.scrollId.getOrElse("Query didn't return a scroll id")
-      queue.enqueue(resp.hits.hits: _*)
+      resp.hits.hits.foreach(h => queue.enqueue(h))
       context become ready
       unstashAll()
   }
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -11,22 +11,22 @@ object Build extends AutoPlugin {
 
   object autoImport {
     val org = "com.sksamuel.elastic4s"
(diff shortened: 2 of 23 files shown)


elastic4s-http-streams/src/main/scala/com/sksamuel/elastic4s/streams/ScrollPublisher.scala
```diff
@@ -150,7 +150,7 @@ class PublishActor(client: HttpClient,
     // more results and we can unleash the beast (stashed requests) and switch back to ready mode
     case Success(resp: SearchResponse) =>
       scrollId = resp.scrollId.getOrElse("Query didn't return a scroll id")
-      queue.enqueue(resp.hits.hits: _*)
+      resp.hits.hits.foreach(h => queue.enqueue(h))
```
samzilverberg, line 153: could be niced with `queue.enqueueAll` and `scala-collection-compat` lib.
I tried it but its missing support for `mutable.Queue.enqueueAll`.
I added it in a PR to the compat lib, but might take a couple of weeks until it appears in a release.

project/Build.scala
```diff
@@ -11,22 +11,22 @@ object Build extends AutoPlugin {
 
   object autoImport {
     val org = "com.sksamuel.elastic4s"
-    val AkkaVersion = "2.4.17"
-    val CatsVersion = "0.9.0"
-    val CirceVersion = "0.7.1"
+    val AkkaVersion = "2.5.27"
+    val CatsVersion = "2.0.0"
+    val CirceVersion = "0.12.0-M3"
```
samzilverberg, line 16: another pr adding 2.13 support to 6.5 handles this differently:
https://github.com/sksamuel/elastic4s/pull/1874/files (search for circeversion)

I could try doing the same if you think its important.

sksamuel: This is fine.

sksamuel:
I've released 5.611

Issue #1956, opened 2019-12-05
getMapping returns only properties but not meta fields


an-tex:
It's possible to retrieve the meta fields using `getIndex(...).mappings.meta`, but shouldn't `getMapping` return those fields as well?

The `getMapping` handler returns only the `properties`:
https://github.com/sksamuel/elastic4s/blob/13f1810502429a2788e1c95f2de15cc32c41f425/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/indexes/MappingHandlers.scala#L22

Reference: https://www.elastic.co/guide/en/elasticsearch/reference/7.5/mapping-fields.html

an-tex:
Hey @sksamuel , I'd be happy to contribute. I just wanted to clarify beforehand if this makes sense. Cheers

sksamuel:
Yes it makes sense to return the fields as well.

PR #2082, opened 2020-04-30
getMappings should return meta

an-tex:
fixes https://github.com/sksamuel/elastic4s/issues/1956


elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/indexes/MappingHandlers.scala
```diff
@@ -3,7 +3,7 @@ package com.sksamuel.elastic4s.requests.indexes
 import com.sksamuel.elastic4s._
 import com.sksamuel.elastic4s.requests.mappings.{GetFieldMappingRequest, GetMappingRequest, PutMappingRequest}
 
-case class IndexMappings(index: String, mappings: Map[String, Any])
+case class IndexMappings(index: String, mappings: Map[String, Any], meta: Map[String, Any] = Map.empty[String, Any])
```
an-tex, line 6: Is it ok to give the meta field in `IndexMappings` a default value? I thought this way it wouldn't break existing code?

sksamuel: Yes makes sense.

Issue #2081, opened 2020-04-30
Elastic container in docker compose used in tests uses default 9300 port


an-tex:
The docker-compose configuration uses the default 9300 elastic search ports to provide an instance for the tests. In case you already have another elastic search instance running locally, you need to stop it and start it again once you're done doing any contribution work on this. It's not a biggy, just an inconvenience which might easily be fixed by changing the ports to something more random but still es-likish like 9337 in both, docker-compose and [DockerTests.scala](https://github.com/sksamuel/elastic4s/blob/master/elastic4s-testkit/src/main/scala/com/sksamuel/elastic4s/testkit/DockerTests.scala )? 
Well and in case you forget to stop your existing instance the tests write in your existing database if it's just a default container without authentication ;)

I'd be happy to PR this if this makes sense

sksamuel:
Yes that makes total sense. We can use an obscure port.

PR #2083, opened 2020-04-30
use non-default elastic ports for tests and docker-compose

an-tex:
fixes https://github.com/sksamuel/elastic4s/issues/2081

I've changed the ports to 
```
      - 39227:9200
      - 39337:9300
```
which are free according to IANA


sksamuel:
Nice.

PR #2084, opened 2020-05-03
Added comprehensive docs on analysis

sksamuel:
Building on your work @crockpotveggies.


crockpotveggies:
Nice! Yes this picks up some of the analysis API that I was not able to capture.

PR #2098, opened 2020-05-13
Update list of passed arguments

Sobolvitya:
Hello guys!

This is small change in documentation. 

`JavaClient` doesn't have predefined parameters in the constructor so we have to specify requestConfigCallback argument.


sksamuel:
I added that in master but it's not released yet. So good shout.

PR #2119, opened 2020-06-01
Add histogram field

Philippus:
Applying this PR will add the histogram field, which was introduced in Elasticsearch 7.6 (https://www.elastic.co/guide/en/elasticsearch/reference/7.7/histogram.html).


sksamuel:
Lgtm

PR #2133, opened 2020-06-16
make seed and field optional to more closely mimic ES api

### LINKED ISSUE #2131: `RandomScoreFunction` not compatible with < ES6.0
ES 6.0 changed the API to require `field` if providing `seed` (https://www.elastic.co/guide/en/elasticsearch/reference/6.0/query-dsl-function-score-query.html#function-random), and your library supports that perfectly

However, the API previous to 6.0, such as 5.6, only supports `seed`, and doesn't support a custom `field`: https://www.elastic.co/guide/en/elasticsearch/reference/5.6/query-dsl-function-score-query.html#function-random
Attempting to pass in a value for `field` produces the following 400: 
```
{"error":{"root_cause":[{"type":"parsing_exception","reason":"random_score query does not support [field]","line":1,"col":193}],"type":"parsing_exception","reason":"random_score query does not support [field]","line":1,"col":193},"status":400}
```

Would you be open to changing
```scala
case class RandomScoreFunction(seed: Long, fieldName: String = "_seq_no", weight: Option[Double] = None, override val filter: Option[Query] = None)
``` 
to something like
```scala
case class SeedAndField(seed: Long, fieldName: String = "_seq_no")
case class RandomScoreFunction(seedAndField: Option[SeedAndField] = None, weight: Option[Double] = None, override val filter: Option[Query] = None)
```

This would allow users to call `RandomScoreFunction()` on older versions of <ES6, while still providing correct behavior on >=ES6. It won't allow older versions to use `seed`, but that's ok for my use case 😅 

Happy to provide a PR if you think this is acceptable. If not... is there any other way around the issue (other than upgrading ES versions...)? I would love to just override `RandomScoreFunctionBuilderFn`, but that's not possible....

dontgitit:
fixes #2131 

changes seed+field pair to be optional

### DIFF
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/searches/queries/funcscorer/RandomScoreFunction.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/searches/queries/funcscorer/RandomScoreFunction.scala
@@ -3,10 +3,17 @@ package com.sksamuel.elastic4s.requests.searches.queries.funcscorer
 import com.sksamuel.elastic4s.requests.searches.queries.Query
 import com.sksamuel.exts.OptionImplicits._
 
-case class RandomScoreFunction(seed: Long, fieldName: String = "_seq_no", weight: Option[Double] = None, override val filter: Option[Query] = None)
+case class SeedAndField(seed: Long, fieldName: String = SeedAndField.DefaultFieldName)
+object SeedAndField {
+  final val DefaultFieldName = "_seq_no"
+}
+case class RandomScoreFunction(seedAndField: Option[SeedAndField], weight: Option[Double], override val filter: Option[Query])
     extends ScoreFunction {
 
-  def fieldName(fieldName: String): RandomScoreFunction = copy(fieldName = fieldName)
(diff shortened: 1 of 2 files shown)


elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/searches/queries/funcscorer/RandomScoreFunction.scala
```diff
@@ -3,10 +3,17 @@ package com.sksamuel.elastic4s.requests.searches.queries.funcscorer
...
-case class RandomScoreFunction(seed: Long, fieldName: String = "_seq_no", weight: Option[Double] = None, override val filter: Option[Query] = None)
+case class SeedAndField(seed: Long, fieldName: String = SeedAndField.DefaultFieldName)
+object SeedAndField {
+  final val DefaultFieldName = "_seq_no"
+}
+case class RandomScoreFunction(seedAndField: Option[SeedAndField], weight: Option[Double], override val filter: Option[Query])
     extends ScoreFunction {
 
-  def fieldName(fieldName: String): RandomScoreFunction = copy(fieldName = fieldName)
+  def fieldName(fieldName: String): RandomScoreFunction = copy(seedAndField = Some(seedAndField.getOrElse(SeedAndField(seed = System.nanoTime())).copy(fieldName = fieldName)))
```
dontgitit, line 13: didn't know what to use for a default seed so used current time...

sksamuel:
Thanks!

Issue #2127, opened 2020-06-04
Include _source options when executing bulkly updates


pgorin:
Hi, Stephen, how are you!
I'm currently working in order to incoporate this options:
**_source
_source_includes
_source_excludes**

when building bulk update requests.

Also I will have to refactor BulkItemResponse in order to, also as UpdateResponse, contain optionally the UpdateGet response from ES. Here I have a question that I will be glad to share with you. We can hold this option get, private, directly on BulkItemResponse class, knowing that this class is also used for Index and Delete bulk update requests, and fetching the source is not available for this 2 operations. But, if this get response is optional, there will be no problem.

Maybe another more elegant solution is to convert BulkItemResponse to a trait or an abstract class and create IndexBulkResponse, UpdateBulkResponse and DeleteBulkResponse.

Again, I will be glad if you share with me your toughs about this modelling issue.
Regards,
Patricio.

sksamuel:
I think the second option is best - have BulkItemResponse as a trait with 3 subtraits, IndexBulkResponse, UpdateBulkResponse and DeleteBulkResponse

pgorin:
Great, I will take this approach. Thank you!

PR #2143, opened 2020-06-22
Supported source fetching when executing bulk update requests.

pgorin:
Feature #2127


sksamuel:
Because this is a breaking change I'd rather include it with a minor point release than a bug release. So I'll hold the PR for 7.9.

pgorin:
Great Stephen, I'll wait for the 7.9 release.
Thank you a lot.

sksamuel:
We can merge this in now as I prepare the 7.9.0 release.

PR #2147, opened 2020-06-24
parse nested field properties in GetIndexResponse

hughsimpson:
ES can returned nested property descriptions, which we weren't parsing. In those nested cases, the `type` field may also be absent (it may _also_ be present, e.g. if the field is a 'nested', rather than 'object' type)


sksamuel:
:+1:

PR #2173, opened 2020-07-18
Fix deprecation warnings for nested sorting

TactiC:
The `nested_filter` and `nested_path` fields have been deprecated since 6.1
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html#nested-sorting


sksamuel:
:+1:

PR #2178, opened 2020-07-25
Fix field sort when both nested path and nested filter are defined

Philippus:
Applying this PR will fix the field sort when both nested path and nested filter are defined. Currently when a nested filter is defined it will override the nested path in the builder.


sksamuel:
:+1:

PR #2186, opened 2020-08-03
backport CompletionSuggestionBuilderFn regex problem from 7.8.1 to 7.5.1

samzilverberg:
https://github.com/sksamuel/elastic4s/pull/2152

backported that simple fix to 7.5.x(via cherrypick) because thats the version of ES i'm stuck with atm 🤷‍♂️ 

if there's any other way you prefer doing backports let me know


samzilverberg:
the travis fail seems rather random.
its failing at some test thats not related at all to this simple `???` removal.

sksamuel:
Does seem unrelated.

PR #2146, opened 2020-06-24
Support basic authentication in the akka-http client

mrooding:
# Context

Supports basic authentication for ElasticSearch with the akka-http client: #2145

# Note

My IDE that's set up with `scalafmt` reformatted the files I touched. I couldn't find any information on which formatter is used in elastic4s so please let me know how you'd like me to format the code to be in sync with the rest.

PR #2195, opened 2020-08-17
Fix setting Akka HTTP credentials

wuservices:
Fixes a bug from #2146 which caused the credentials to be effectively ignored. Since `httpRequest.addCredentials` returns a copy of `httpRequest` the original implementation didn't work. Now, the copy with the credentials is used when making the request.


sksamuel:
Nice spot.

PR #2217, opened 2020-09-18
Add wildcard field

Philippus:
Applying this PR will add the wildcard field, which was introduced in Elasticsearch 7.9 (https://www.elastic.co/guide/en/elasticsearch/reference/7.9/keyword.html#wildcard-field-type).


sksamuel:
:+1:

PR #2239, opened 2020-10-26
Add Twitter to list of adopters

meitemarkymark:
Twitter is a Scala shop and has adopted this library for some internal customers.


sksamuel:
:+1:

Issue #2241, opened 2020-10-27
Script parameters should support case classes


andrei-micu:
When sending case classes as parameter values to a script, there is no complaining from the elastic4s API, but it can result in cryptic errors coming from the Elasticsearch cluster. Ex:
```scala
case class A(foo: String, bar: Int)

val aInstance = A("a", 1)

val request = updateById(indexName, documentId)
       .script {
         script(s"ctx._source.a=params.a;")
           .params("a" -> aInstance)
       }

client.execute {
  request
}... 
// Will result in a failure with this nested exception:
// ElasticError(mapper_parsing_exception,object mapping for [a] tried to parse field [null] as object, but found a concrete value,...)
```

From my research, mostly based on what I saw in the `autofield` and `autovalue` methods here https://github.com/sksamuel/elastic4s/blob/f56364c2758bab20385fb49fc1a31315dc9da53c/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/json/XContentBuilder.scala#L213 , I concluded that the parameters were thought to be populated by Java/Scala primitives and collections, but not by case classes.
The addition of logic to handle `Product` instances as arrays here https://github.com/sksamuel/elastic4s/commit/b3c1dc76dbfd71333090bf21cc490860979f1d93#diff-92abf159bb196014f430507d73cce08dcbe2d842b4817df0a5d53ca609025865R186 lead to serializing case classes as arrays as well, which in turn lead to the types of errors described above (in which arrays were assigned to fields that had an object mapping).

What I propose is to deprecate the `param`/`params` methods in https://github.com/sksamuel/elastic4s/blob/master/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/script/Script.scala and to add the following set of methods (along with their versions for adding multiple parameters at once):
- `paramPrimitive(name: String, value: Any)` - basically what the previous `param` method did, but now it states it explicitly;
- `paramObject(name: String, value: Any)(implicit paramMarshaler: ParamMarshaler[T])`
  - takes an object as parameter value and serializes it based on the supplied (implicit) marshaler;
  - the ParamMarshaler will be a new elastic4s implicit converter in the same spirit as the `Indexable`s (https://github.com/sksamuel/elastic4s/blob/master/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/Indexable.scala).
- `paramRaw(name: String, json: String)` - this one will take string values representing the parameter serialized as JSON (useful for people that don't like implicits).

[...]

PR #2244, opened 2020-11-02
Added script parameter object values functionality.

andrei-micu:
Implemented the proposed changes in https://github.com/sksamuel/elastic4s/issues/2241 with some slight changes based on feedback I received.
Also added the SprayJson version of the ParamSerializer derivation logic.


sksamuel:
Thanks, will review this in the next few days!

sksamuel (approved):
awesome

PR #2252, opened 2020-11-12
fix: source is ignored when using `multi()`

brodin:
When invoking the `multi` search anything on request.source is ignored since this logic lives in the SearchHandlers.scala


sksamuel (approved):

brodin:
@sksamuel how come you reverted my changes here https://github.com/sksamuel/elastic4s/commit/a060a0a1e7b790eca725cdf6211d267bae2cf238 ? :)

sksamuel:
Looks like it was failing the build for some reason @brodin

PR #2267, opened 2020-12-07
[6.7.x] Fix deprecated source params and add `retry_on_conflict` to update request in bulk

livelxw:
Fix warning `Deprecated parameter [_source_include] used, expected [_source_includes] instead`.
Add `retry_on_conflict` to update request in bulk


sksamuel:
:+1:

Issue #2272, opened 2020-12-12
AkkaHttpClient: java.lang.IllegalStateException: Promise already completed


therockstorm:
In the event of a failure, [this code block](https://github.com/sksamuel/elastic4s/blob/35ae2b8fb2986472559dbcf554b83997b3861d1d/elastic4s-client-akka/src/main/scala/com/sksamuel/elastic4s/akka/AkkaHttpClient.scala#L47-L57) causes a `java.lang.IllegalStateException: Promise already completed` exception because the promise is `s.host.success(host)` completes the promise and then `s.host.failure(e)` tries to complete it again. This can be fixed by moving `s.host.success(host)` into the `case Success(req)` block.

PR #2273, opened 2020-12-12
Move s.host.success(host) into Success case to avoid 'Promise already completed' exception

therockstorm:
Fixes https://github.com/sksamuel/elastic4s/issues/2272


sksamuel:
:+1:

PR #2280, opened 2020-12-22
Fixes #2238: upsert UpdateRequest strips empty docs

### LINKED ISSUE #2238: upsert UpdateRequest strips empty upsert docs
https://github.com/sksamuel/elastic4s/blob/master/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/update/UpdateBuilderFn.scala#L33

It's valid in ES to have an empty upsert doc (i.e. no document field updates), if one is providing a `script`. This is especially useful in scripted upserts, where the script is always run. However, it's also valid to simply choose to use painless scripting to construct the upserted document.

In either scenario, an empty `upsert` doc is needed to tell ES that the update is an upsert. The `UpdateBuilderFn` is stripping the upsert marker if the doc is empty.

A valid request that fails with e4s might look like this:
```
{
 "script": {
    "source": "...",
    "lang": "painless"
  },
  "upsert": {}
}
```

I would recommend simply removing the `nonEmpty` guard around this JSON rendering such that `"upsert": {}` is preserved. Our workaround for now is to simply add a dummy field in the `upsert` document.

tarangbhalodia:


sksamuel:
Thanks.

Can you check the tests.

tarang-vw:
@sksamuel Looks like a configuration error for the failing tests. Do you see any other issues?

tarangbhalodia:
@sksamuel  Looks like a configuration error for the failing tests. Do you see any other issues?

sksamuel:
> @sksamuel Looks like a build configuration error for the failing tests. Do you see any issue with the current fix that can cause the failures?

Looks like you fixed it before I had chance to look :)

PR #2308, opened 2021-02-03
Add version field

Philippus:
Applying this PR will add the version field, which was introduced in Elasticsearch 7.10 (https://www.elastic.co/guide/en/elasticsearch/reference/7.10/version.html).


sksamuel:
Great, will release another 7.10 version with this.

PR #2328, opened 2021-02-23
fix for targetType in reindex request

### LINKED ISSUE #2326: ReindexRequest.targetType applied to source index iso target index
The `targetType` field of the `ReindexRequest` is applied to the `source.index` object of the actual http request
but it should be applied to the `dest.index`

see https://github.com/sksamuel/elastic4s/blob/78a4c384376393f9ea69d16501d7de805c7b0f9e/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/reindex/ReindexBuilderFn.scala#L18-L40

brunoballekens:
fixes #2326

I fixed it on master because I could not find the ReindexTest on the 6.7.x branch.
The test `apply targetType to dest index - not to the source index` however does fail if the fix is not applied - suggesting that,  contrary to what is specified in the elasticsearch documentation about the reindex, the type is applied also on the source if specified 🤷 .


sksamuel:
Looks good to me.

Issue #2349, opened 2021-03-12
NPE when reading stats aggregation result when value is null


epeden:
Empty Stats aggregation fails with a null pointer exception. Related to [1614](https://github.com/sksamuel/elastic4s/issues/1614), but in this situation the count is 0 and "min", "max", and "avg" keys exist but with null values. So when those keys are accessed and converted toString, NPE is thrown.
```
java.lang.NullPointerException: null
	at com.sksamuel.elastic4s.requests.searches.aggs.responses.HasAggregations.statsBucket(aggresponses.scala:308)
	at com.sksamuel.elastic4s.requests.searches.aggs.responses.HasAggregations.statsBucket$(aggresponses.scala:304)
	at com.sksamuel.elastic4s.requests.searches.aggs.responses.Aggregations.statsBucket(aggresponses.scala:146)
```

PR #2350, opened 2021-03-12
Wrap StatsBucketAggResult values in Option before converting

epeden:
Bug: [2349](https://github.com/sksamuel/elastic4s/issues/2349). When the StatsBucketAggResult is empty, the "count" and "sum" are 0, but the values for the keys "min", "max", and "avg" are null, so a Null Pointer Exception is thrown when converting toString.toDouble. By wrapping these in an Option, we can check that a value exists and set to 0 if not.


sksamuel:
Thanks.

PR #2353, opened 2021-03-18
Adds case_insensitive flag to Wildcard query

jmaasing:
Adds support for case_insensitive flag introduced in ES 7.10.0

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-wildcard-query.html#wildcard-query-field-params

I'm not sure what needs to be done to support the DSL since I don't use that myself.


sksamuel:
That's all we need for now, looks good!

Issue #2352, opened 2021-03-16
URL encoding change


ievalu:
Hi,

there was a breaking change introduced in ElasticSearch v7.4. The plus sign encoding in URLs changed. 
It seems like in elastic4s `UrlEncoder` is used which is not universal, i.e. applicable only for a specific content type.
It affects ES `_id`s which have spaces in them because `UrlEncoder` encodes spaces as plus signs while spaces should be encoded as `%20`. 

_Example situation_:
Indexing document with `_id` = `test id`. The actual document's `_id` in ElasticSearch index will be `test+id`. The expected behaviour is for `_id` to be `test id`.

Refs: 
https://www.elastic.co/guide/en/elasticsearch/reference/7.x/breaking-changes-7.4.html#_changes_to_encoding_plus_signs_in_urls
https://docs.oracle.com/javase/8/docs/api/java/net/URLEncoder.html

ievalu:
Could I work on this issue?

sksamuel:
Of course.

PR #2356, opened 2021-03-22
Fix URL encoding of spaces

ievalu:
References https://github.com/sksamuel/elastic4s/issues/2352

I chose to just replace all plus signs with `%20`. Plus signs are encoded to `%2B` so this shouldn't break anything.


sksamuel:
Good work thanks

ievalu:
@sksamuel Could you release it?

sksamuel:
I can do a 7.4 release yeah.

sksamuel:
7.4.1 is out, takes ~hour to sync with central

ievalu:
@sksamuel Hello once again, I've tried to use 7.4.1v in our project and it seems like it was released incorrectly, maybe you could check it out https://github.com/sksamuel/elastic4s/compare/v7.4.0...v7.4.1

PR #2357, opened 2021-03-23
Fix UpdateById upsert object auto-inserting

### LINKED ISSUE #2344: updateById with script always upserts
Hello, due to fix for #2238 ([commit](https://github.com/sksamuel/elastic4s/commit/307777b1b331a9f9f5acd40e051456f626fe7254#diff-71e19d3f006416bcee1a014e3517de72511bb4fd46177f831c3a3759e9df1f5dR33)), whenever an `updateById` query contains a script, it will always work like an upsert and will create empty document if it is not present.

This may be undesirable, and can't be prevented in any way using the elastic4s API. Perhaps tie the condition to `&& scripted_upsert == true` as well?

Tasssadar:
Due to 307777b1b331a9f9f5acd40e051456f626fe7254, it was impossible
to create non-upsert update query with script.

DocAsUpsert also no longer inserts empty upsert object into the query,
in line with ElasticSearch documentation.

Fixes #2344

### DIFF
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/update/UpdateBuilderFn.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/update/UpdateBuilderFn.scala
@@ -30,7 +30,7 @@ object UpdateBuilderFn {
       builder.rawField("upsert", upsert)
     }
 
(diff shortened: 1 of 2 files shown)


Tasssadar:
The test failed, but not because of anything in this PR.

sksamuel:
Looks good, thanks for the fix.

PR #2372, opened 2021-03-30
Add ParamSerializer[T] instances for circe, play, json4s and jackson

charlescd:
The implicit to derive a ParamSerializer[T] from a json serializer was only implemented for spray json.

I thought it would be a good idea to add one for each json library supported by elastic4s.


sksamuel:
Nice job.

PR #2383, opened 2021-04-09
Update ip field

Philippus:
Applying this PR will update the `ip` field. As far as I can see, reading https://www.elastic.co/guide/en/elasticsearch/reference/current/ip.html, the `copyTo` and `properties` are not valid parameters but `ignoreMalformed` and `nullValue` are.


sksamuel:
👍🏼

PR #2390, opened 2021-04-17
Implement async search

charlescd:
This is a PR to handle [async search](https://www.elastic.co/guide/en/elasticsearch/reference/current/async-search.html).

I prefer to create it as draft for now because IMO: 
 - It lacks tests and documentation.
 - I implemented it using a debatable approach and I want to be sure it is ok before going any further.

As the [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/async-search.html) says, async search supports the same parameters as search. Thus, to avoid duplicate code, this implementation use `SearchRequest` directly. The downside is that as it is indicated, scroll is not supported. But with this implementation, one can set a (useless) scroll parameter which is not "clean".


elastic4s-core/src/main/scala/com/sksamuel/elastic4s/api/AsyncSearchApi.scala
```diff
@@ -0,0 +1,15 @@
+package com.sksamuel.elastic4s.api
+
+import com.sksamuel.elastic4s.requests.searches.{AsyncSearchRequest, AsyncSearchStatusRequest, DeleteAsyncSearchRequest, FetchAsyncSearchRequest, SearchRequest}
+
+trait AsyncSearchApi {
```
sksamuel, line 5: I wouldn't add this. I'm moving towards creating the case classes directly.

sksamuel (changes requested):

sksamuel:
I think the approach is fine, if there is only that one parameter difference.
Otherwise you could look at the way mapping definition and put mapping work. There's not a lot of re-use that we can do in these situations - copy n paste or reuse.

PR #2393, opened 2021-04-20
Fix auto_date_histogram object name

jamieparkinson:
Looks like this was probably a copy-paste error - the auto date histogram added in https://github.com/sksamuel/elastic4s/commit/49ff096ea89058d7acbafbb1ab8ba453564d8fa3 should have been named `auto_date_histogram` https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-autodatehistogram-aggregation.html


sksamuel:
👍🏼

PR #2408, opened 2021-04-30
Added support for None type

goldlil:
Currently None type will match Product type and thus type will be empty array instead of null


sksamuel:
Thanks.

PR #2410, opened 2021-04-30
Update README.md

emret:
For multiple endpoints example, url schema prefix should be exists only for the first endpoint.


sksamuel:
Thank You

sksamuel:
👍🏼

PR #2412, opened 2021-04-30
Update create request match with None

goldlil:


goldlil:
@sksamuel not sure if this is right branch to create PR, but I need this bug fix for this version of elastic. Array instead of null is creating us a lot of problem.

sksamuel:
Yes this is fine, I will release a 7.3.x version with it in.

PR #2425, opened 2021-05-18
Fix deserialization of some elastic responses to snapshot operations

hughsimpson:
'default' `RestoreSnapshotResponse` contains an `accepted` boolean, not `acknowledged`. Also, if waitForCompletion = true, the response is in a different format entirely. This fixes response deserialization for `RestoreSnapshotResponse` and `CreateSnapshotResponse`, but it's highly probably that there are many other cases where the response from elasticsearch doesn't match the response type from the handler... Haven't got into any further depth on that possibility here, though.

### DIFF
--- a/elastic4s-domain/src/main/scala/com/sksamuel/elastic4s/requests/snapshots/response.scala
+++ b/elastic4s-domain/src/main/scala/com/sksamuel/elastic4s/requests/snapshots/response.scala
@@ -5,7 +5,7 @@ import com.sksamuel.elastic4s.requests.common.Shards
 import scala.concurrent.duration._
 
 case class CreateRepositoryResponse(acknowledged: Boolean)
-@JsonTypeInfo(use = JsonTypeInfo.Id.DEDUCTION, property = "accepted")
+@JsonTypeInfo(use = JsonTypeInfo.Id.DEDUCTION)
 @JsonSubTypes(
   Array(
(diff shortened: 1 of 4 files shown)


hughsimpson:
Reopening now that tests pass and more cases work

elastic4s-domain/src/main/scala/com/sksamuel/elastic4s/requests/snapshots/response.scala
```diff
@@ -1,11 +1,22 @@
...
 
 case class CreateRepositoryResponse(acknowledged: Boolean)
-case class CreateSnapshotResponse(accepted: Boolean)
+@JsonTypeInfo(use = JsonTypeInfo.Id.DEDUCTION)
+@JsonSubTypes(
+  Array(
+    new JsonSubTypes.Type(value = classOf[CreateSnapshotResponseAsync], name = "accepted"),
+    new JsonSubTypes.Type(value = classOf[CreateSnapshotResponseAwait], name = "snapshot")))
+sealed trait CreateSnapshotResponse { def succeeded: Boolean }
+case class CreateSnapshotResponseAsync(accepted: Boolean) extends CreateSnapshotResponse {
```
hughsimpson, line 14: Naming could probably be improved here 😅 I'm open to suggestions

sksamuel: I think these names are fine, they're clear.

PR #2429, opened 2021-05-23
Add aggregate-metric field

Philippus:
Applying this PR will add the aggregate-metric field, which was introduced in Elasticsearch 7.11 (https://www.elastic.co/guide/en/elasticsearch/reference/7.11/aggregate-metric-double.html#aggregate-metric-double).


sksamuel:
Great 👍🏼

PR #2430, opened 2021-05-24
Uncomment and fix dynamic templates tests

hughsimpson:
This also adds a DynamicType field, to support one of the commented-out tests, although I'm unfamiliar with its usage, so the fields I've made available for it may not be the most appropriate. There are other templates that this leaves unsupported (e.g. https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic-templates.html#dynamic-mapping-runtime-fields - with which I am also unfamiliar) but this seems to fix some common use cases...

Couple of unrelated things:
- fix the locale field name in DateFieldBuilderFns
- Remove duplicate setting of fielddata in TextFieldBuilderFn


sksamuel:
Awesome thanks.

PR #2447, opened 2021-06-08
Add support for reindex conflicts field

dragisak:
As described in the [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-reindex.html#docs-reindex-api-request-body) `conflicts` field is part of the request body.


sksamuel:
Awesome thanks. Will release.

PR #2449, opened 2021-06-09
sbt cleanup

zafarella:
* move deps to separate file
* address some of the sbt warnings on outdated syntax
* use G1 GC for sbt run
* remove "sbt-dotty" plug-in


sksamuel:
Very neat.

sksamuel:
Thanks :)

Issue #2379, opened 2021-04-05
May I propose a more pure FP design for `ElasticClient`, `HttpClient` and `Executor`?


kailuowang:
The current design tries to abstract the effect type `F[_]` into `Executor[F]`, but in reality, the effect type is usually determined by the implementation in `HttpClient`. For example, `SttpRequestHttpClient` is actually using a `Future`... 
May I propose simplifing this design by using a barebone tagless final design and rid of the need for   `Exceutor`. 
Bascially 
```Scala
trait HttpClient[F[_]] {
  def send(request: ElasticRequest): F[HttpResponse] 
}

object Elastic4s {

    def execute[T, U, F[_]](t: T)(implicit
                                    client: HttpClient[F],
                                    functor: Functor[F],
                                    handler: Handler[T, U],
                                    options: CommonRequestOptions): F[Response[U]] {
          val request = handler.build(t)         
          //omitting the part that add `timeout` and `master_timeout` parameter
          functor.map(client(request3)) { resp =>
              handler.responseHandler.handle(resp)
                .fold(
                  error =>
                    RequestFailure(resp.statusCode, resp.entity.map(_.content), resp.headers, error),
                  u =>
                    RequestSuccess(resp.statusCode, resp.entity.map(_.content), resp.headers, u)
                )
     }
}

```

With this design it's quite easy to add more `HttpClient` implementations, specifically, it made it a lot easier for each `HttpClient` implementation to control their effect type system. The current `ElasticClient` is a simple wrapper of `HttpClient`, I feel it's unnecessary for it to try to manage the effect system for `HttpClient`

Here is an example of a `Http4s` based `HttpClient` impl.

```scala

class Http4sElastic4sClient[F[_]: ContextShift](
    baseUri: Uri,
    blocker: Blocker,
    hClient: Client[F]
  )(implicit
    F: Effect[F]) {

  private def processResponse(resp: Response[F]): F[HttpResponse] = {
    resp.as[String].map { body =>
      val entity =
        Some(HttpEntity.StringEntity(body, resp.contentType.map(_.renderString)))
      HttpResponse(
        resp.status.code,
        entity,
        resp.headers.toList.map(h => (h.name.value, h.value)).toMap
      )
    }
  }

  implicit val fileEncoder = EntityEncoder.fileEncoder[F](blocker)
  implicit val isEncoder = EntityEncoder.inputStreamEncoder[F, InputStream](blocker)

  private def send(
      request: ElasticRequest
[...]

PR #2459, opened 2021-06-19
Fix Cats Effect IO type inference

yangzai:
As mention in this [comment](https://github.com/sksamuel/elastic4s/pull/1981#issuecomment-702211419) the generalisation of implicits to provide `Executor`/`Functor` typeclass instances for all types with `Async`/`CatsFunctor` instances loses `IO` specificity and implicit precedence priority, making client execution infer to the default `Future` effect return type. This makes using the `IO` effect type very unwieldy, having to explicitly state all type parameters for client execution.

This PR brings back `IO` type inference without losing support for other `Async` effect type.


elastic4s-effect-cats-2/src/main/scala/com/sksamuel/elastic4s/cats/effect/instances/CatsEffectInstances.scala
```diff
@@ -12,6 +12,9 @@ trait CatsEffectInstances {
 
   implicit def catsEffectExecutor[F[_]: Async]: Executor[F] =
     new CatsEffectExecutor[F]
+
+  //this needs to be at the bottom
```
yangzai, line 16: I don't know why the ordering matters.

An alternative implementation with a lower priority trait `CatsEffectInstances0` also works:
```scala
trait CatsEffectInstances extends CatsEffectInstances0 {
  implicit def catsFunctor[F[_]: CatsFunctor]: Functor[F] = new Functor[F] {
    override def map[A, B](fa: F[A])(f: A => B): F[B] = CatsFunctor[F].map(fa)(f)
  }

  implicit def catsEffectExecutor[F[_]: Async]: Executor[F] =
    new CatsEffectExecutor[F]
}

trait CatsEffectInstances0 {
  implicit val ioExecutor: Executor[IO] = new CatsEffectExecutor[IO]
}
```
but I'm equally confused as to why the more specific `ioExecutor` needs to be in the lower and not the higher priority trait for it to work.

sksamuel: I  want to do https://github.com/sksamuel/elastic4s/issues/2379 as well for 8.0. Maybe that's easier to do now ?

yangzai: > I want to do #2379 as well for 8.0. Maybe that's easier to do now ?

I think they are mutually exclusive. For #2379 most of the work seem to revolve around refactoring `HttpClient`.

The main pain points with the current `execute` for me are:
- Defaulting to `Future`
The default `Future` typeclass instances in the companion objects for `Functor`/`Executor` seems conflict prone to me in terms of implicit resolution. Maybe you could drop the default instances and have 2 execution methods instead (`execute`/`executeF`) where one always return a `Future`. Or just keep to 1 and have users manually import `Future` instances.
2. The type parameters `[T, U, F[_]]`
Having to be explicit with the effect type `F` is fine but it forces you to have to figure out what `T` and `U` are. Being able to explicit on `F` without `T` and `U` would be nice, perhaps something in the direction of `client.effect[IO].execute(...)`

sksamuel:
This is a little beyond my knowledge. What's the tl;dr of the change ?

elastic4s-effect-cats-2/src/test/scala/com.sksamuel.elastic4s.cats.effect/CatsEffectTest.scala
```diff
@@ -0,0 +1,36 @@
...
+import cats.effect.IO
+import com.sksamuel.elastic4s.{ElasticClient, ElasticProperties}
+import com.sksamuel.elastic4s.ElasticDsl._
+import org.scalatest.flatspec.AnyFlatSpec
+import com.sksamuel.elastic4s.cats.effect.instances._
+import com.sksamuel.elastic4s.http.JavaClient
+import com.sksamuel.elastic4s.requests.indexes.admin.DeleteIndexResponse
+import com.sksamuel.elastic4s.requests.indexes.{CreateIndexRequest, CreateIndexResponse, DeleteIndexRequest}
+
+class CatsEffectTest extends AnyFlatSpec {
```
yangzai, line 13: @sksamuel basically, without the changes in `CatsEffectInstances` the first test case would not compile unless we rewrite it with explicit type params like the second test case.

sksamuel: Do users of standard scala futures have any breaking changes ?

yangzai: If users import `com.sksamuel.elastic4s.cats.effect.instances._` they intend `IO` to be the default effect. `Futures` should still work with explicit type params even with Cats Effect instances imported. I can add another test for that.

sksamuel: Ok great. I just wanted to confirm that people who are not explicitly using effects would still be getting the default Futures implementation.

yangzai: @sksamuel Test cases updated and branch rebased. I will need approval to trigger CI.

sksamuel:
Triggered

PR #2472, opened 2021-07-16
Add task support to DeleteByQuery in 7.9

izzyreal:
We add `wait_for_completion` support to `Delete`.

To not break the 7.9 `Response` API, we don't touch `DeleteByQueryHandler`. Instead, to get the `taskId` of a `DeleteByQuery` that does not `wait_for_completion`, the consuming side can define a custom `DeleteByQueryHandler` and do the below:
```
def deleteUserByName(name: String) {
    implicit val handler: Handler[DeleteByQueryRequest, Either[DeleteByQueryResponse, CreateTaskResponse]] =
      DeleteByQueryHandlerWithTask
    client
      .execute {
        ElasticDsl.deleteIn("users").by(termQuery("name", name)).waitForCompletion(false)
      }
      ...
```

Special caution must be taken not to have the default `elastic4s` `DeleteByQueryHandler` implicitly in scope, for example:
```
import com.sksamuel.elastic4s.ElasticDsl.{ DeleteByQueryHandler => _, _ }
```

### DIFF
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/delete/DeleteByQueryRequest.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/delete/DeleteByQueryRequest.scala
@@ -14,6 +14,7 @@ case class DeleteByQueryRequest(indexes: Indexes,
                                 proceedOnConflicts: Option[Boolean] = None,
                                 refresh: Option[RefreshPolicy] = None,
                                 waitForActiveShards: Option[Int] = None,
+                                waitForCompletion: Option[Boolean] = None,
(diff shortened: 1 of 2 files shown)


izzyreal:
@sksamuel For the project I'm on, we're using 7.9. If this is a welcome change on other branches, let me know and I'll open another PR.

sksamuel:
I can release a 7.9 if you want ?
And if you could add to master, that would be great - we can just change the original object in master.

izzyreal:
@sksamuel That's great news, a 7.9 release would be awesome indeed! I'll adjust the `master` PR right now and ping you in that PR when done.

sksamuel:
I released 7.9 last night so you should be good.

PR #2477, opened 2021-07-23
Add task support to DeleteByQuery in master

izzyreal:
We add `wait_for_completion` support to `Delete`.

The project I'm on is using 7.9. The same PR but to the `release/7.9.x` branch can be found here: https://github.com/sksamuel/elastic4s/pull/2472

Edit: **The below does not apply anymore** as per @sksamuel's comment below.

To not break the existing `Response` API, we don't touch `DeleteByQueryHandler`. Instead, to get the `taskId` of a `DeleteByQuery` that does not `wait_for_completion`, the consuming side can define a custom `DeleteByQueryHandler` and do the below:
```
def deleteUserByName(name: String) {
    implicit val handler: Handler[DeleteByQueryRequest, Either[DeleteByQueryResponse, CreateTaskResponse]] =
      DeleteByQueryHandlerWithTask
    client
      .execute {
        ElasticDsl.deleteIn("users").by(termQuery("name", name)).waitForCompletion(false)
      }
      ...
```

Special caution must be taken not to have the default `elastic4s` `DeleteByQueryHandler` implicitly in scope, for example:
```
import com.sksamuel.elastic4s.ElasticDsl.{ DeleteByQueryHandler => _, _ }
```

### DIFF
--- a/docs/version8migration.md
+++ b/docs/version8migration.md
@@ -21,3 +21,5 @@ removed boost from double field
 match queries renamed to include query suffix
 
 suggestion response classes moved to newpackage
+
+DeleteByQueryRequest returns Either[DeleteByQueryResponse, CreateTaskResponse] instead of DeleteByQueryResponse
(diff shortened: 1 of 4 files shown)


sksamuel:
I replied in the other PR. I think for master I'm ok with us changing the response object.

izzyreal:
* Changed the response in congruence with what is done in `ReindexHandlers`
* Fixed existing `DeleteByQueryTest`
* Added a test to make sure our new feature is working as intended
* Amended migration docs

Let me know if something's off about the styling/formatting. Also we could move `TaskRegex` into a shared place like `Handler.scala`.

sksamuel:
TaskRegex could move yeah - if you wanted to make a PR I would merge.
I merged this one in since it was ready to go.

izzyreal:
`TaskRegex` deduplication PR
https://github.com/sksamuel/elastic4s/pull/2491

PR #2482, opened 2021-07-27
Adding Unsigned Long Field Type

cwgroppe:
Added two new case classes, UnsignedLongField, UnsignedLongStringField. The unsigned_long field type in ES can be indexed in a numeric or string form. The unsigned long doesn't have a named range type. Please let me know if anything else is needed.


cwgroppe:
Yeah I think I made some weird typos after doing a post compile review. Like elastic4s was somehow elastic5s in the package declaration

cwgroppe:
third times a charm

sksamuel:
Maybe 4th :)

PR #2490, opened 2021-07-29
extend AkkaHttpClientSettings to skip ssl certificate verification

an-tex:
useful when self signed certificates are used and the jwk truststore can't be changed

### DIFF
--- a/docs/clients.md
+++ b/docs/clients.md
@@ -55,7 +55,7 @@ To use these, add the appropriate module to your build, and then pass an instanc
 For Akka HTTP, we use `AkkaHttpClient`:
 
 ```scala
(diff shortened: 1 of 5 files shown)


docs/clients.md
```diff
@@ -55,7 +55,7 @@ To use these, add the appropriate module to your build, and then pass an instanc
 For Akka HTTP, we use `AkkaHttpClient`:
 
 ```scala
-val client = ElasticClient(AkkaHttpClient(AkkaHttpClientSettings(List("http://host1:9200"))))
```
an-tex, line 58: in my tests I got an error "invalid input" when the scheme was added here. the https config settings should be used to define the scheme

sksamuel:
👍🏼

PR #2498, opened 2021-08-06
fix for ElasticNodeEndpoint prefix getting ignored

stanislav-chetvertkov:


elastic4s-client-esjava/src/main/scala/com/sksamuel/elastic4s/http/JavaClient.scala
```diff
@@ -128,6 +128,9 @@ object JavaClient extends Logging {
 
     val client = RestClient
       .builder(hosts: _*)
+      .setPathPrefix(
+        props.endpoints.map(_.prefix).collectFirst { case Some(prefix) => prefix }.getOrElse("")
```
stanislav-chetvertkov, line 132: would only work if all of the prefixes are the same (or all absent), would it make sense to fail with an exception otherwise?

sksamuel: I think that's an unusual edge case, but we could fail yes.

stanislav-chetvertkov:
I've noticed some validation errors when setting the path to an empty string and changed the logic a little bit to make sure that all of the prefixes are the same.
Apparently I have problems with executing the tests locally (fails even against the master branch with many `com.sksamuel.elastic4s.http.JavaClientExceptionWrapper: java.net.ConnectException: Connection refused`) that is why I've missed the validation errors initially

PR #2502, opened 2021-08-17
(WIP) Scala 3

hughsimpson:
Had a look at how far we could get supporting scala 3. There's a release candidate for scala-modules-jackson which seems to work, although a lot of the subprojects depend on libs that aren't yet published.

Other issues:
- circe _should_ work, but for some reason the default implicit printer params in `indexableWithCirce` and `paramSerializerWithCirce` aren't being picked up. I can't yet minimise this -- _in general_ implicit defaults still work with scala 3 🤷 
- My sbt-fu isn't good enough to avoid some weird warts in the github workflows (seems like running `sbt ++$scalaVersion $command` runs the command against every aggregated subproject, even when they're not explicitly crossbuilt for that version. There _must_ be a way around this...)
- Some of the other required dependencies for other modules are already available as release candidates / milestones (e.g. zio-json and play-json); other than jackson, they're all pretty independent of each other, so I've left them alone for now until a release version is out.
- It's all dependent on the `exts` lib being crosspublished, which is why I made a [pr for that](https://github.com/sksamuel/exts/pull/27)

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -108,7 +108,7 @@ jobs:
         run: docker run -d -it -p 39227:9200 -p 39337:9300 -e "discovery.type=single-node" -v /home/runner/work/elastic4s/elastic4s/elastic4s-tests/src/test/resources/elasticsearch.yml:/usr/share/elasticsearch/config/elasticsearch.yml docker.elastic.co/elasticsearch/elasticsearch:7.14.0
 
       - name: run tests
-        run: sbt ++3.0.1 `for i in json_builder domain handlers core clientcore clientesjava clientsSniffed cats_effect cats_effect_2 tests testkit jackson json4s; do echo $i/test; done | xargs`
+        run: sbt ++3.0.1 `for i in json_builder domain handlers core clientcore clientesjava clientsSniffed cats_effect cats_effect_2 zio tests testkit jackson json4s; do echo $i/test; done | xargs`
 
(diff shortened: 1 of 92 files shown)


sksamuel:
I was thinking we can release scala 3 to coincide with elastic 8.
That way, any breaking changes required can be done at once.

hughsimpson:
That makes sense to me, although I'll be unable to upgrade to elasticsearch 8 - so I hope the intent is to still release a scala-3 compatible 7.x release, and just align the timings a bit

hughsimpson:
I realise now of course that exts dependency doesn't need to be crosspublished at all, since there are no actual incompatibilities, so I'll take a look at the Circe issue some time in the next few weeks. Biggest blocker seems to be the Akka module's test dependencies on scalaMock, which I don't think I can upgrade because of a few shortcomings in scala 3 macros. I think a pure java mock framework (or else the total avoidance of mocks) is the only way forwards there ..

cornerman:
> seems like running sbt ++$scalaVersion $command runs the command against every aggregated subproject, even when they're not explicitly crossbuilt for that version.

@hughsimpson I had the same issue in another project, there seems to be an issue in sbt (https://github.com/sbt/sbt/issues/3465). You can workaround it by setting the `crossScalaVersions := Nil` globally in your `build.sbt` (see [here](https://github.com/cornerman/colibri/blob/master/build.sbt#L2)).

PR #2519, opened 2021-09-06
Add match-only-text field

Philippus:
Applying this PR will add the match-only-text field, which was introduced in Elasticsearch 7.14 (https://www.elastic.co/guide/en/elasticsearch/reference/7.x/text.html#match-only-text-field-type).


sksamuel:
👍🏼

sksamuel:
Will release 7.14.1

PR #2530, opened 2021-09-24
Add support for case-insensitive prefix queries.

nespera:
Also add a test for prefix query body. This has been available in Elasticsearch since 7.10.0. See https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-prefix-query.html#prefix-query-field-params


sksamuel:
awesome thanks

PR #2531, opened 2021-09-24
Add support for case-insensitive term queries.

nespera:
Also add a test for query body. Another field supported since 7.10.0 see https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-term-query.html#term-field-params


sksamuel:
Thanks!

PR #2533, opened 2021-09-27
Add support for combined_fields query

gertv:
This PR adds support for the `combined_fields` query type that was added in Elasticsearch 7.13


sksamuel:
Looks great.

PR #2536, opened 2021-10-04
Fix wrong import for Explanation

### LINKED ISSUE #2532: Possible wrong `Explanation` import in `SearchHit`
Hello,

Currently (in master but is the same in 7.13/7.14) the `Explanation` imported in `SearchHit` is the one from the package `com.sksamuel.elastic4s.requests.validate` : 

https://github.com/sksamuel/elastic4s/blob/3454148330eb86ee918c4ffd34b66f44f430f4cc/elastic4s-domain/src/main/scala/com/sksamuel/elastic4s/requests/searches/SearchHit.scala#L7

which has this shape: 

https://github.com/sksamuel/elastic4s/blob/8b2624bb80086cff9e79f8ffb171a9a254d39ec5/elastic4s-domain/src/main/scala/com/sksamuel/elastic4s/requests/validate/Explanation.scala#L3

but looking at the ElasticSearch responses, the shape of the `_explanation` field is:

```json
	   "_explanation" : {
          "value" : 1.0,
          "description" : "*:*",
          "details" : [ ]
       }
```

which matches the `Explanation` from the package `com.sksamuel.elastic4s.requests.explain`: 

https://github.com/sksamuel/elastic4s/blob/3454148330eb86ee918c4ffd34b66f44f430f4cc/elastic4s-domain/src/main/scala/com/sksamuel/elastic4s/requests/explain/ExplainResponse.scala#L6

Is maybe the import in `com.sksamuel.elastic4s.requests.searches.SearchHit` incorrect and should be

```scala
import com.sksamuel.elastic4s.requests.explain.Explanation
```
instead of 
```scala
import com.sksamuel.elastic4s.requests.validate.Explanation
```
?

Changed it locally and compiled just fine, tests also compile correctly.

Dbzman:
This should fix #2532

### DIFF
--- a/elastic4s-domain/src/main/scala/com/sksamuel/elastic4s/requests/searches/SearchHit.scala
+++ b/elastic4s-domain/src/main/scala/com/sksamuel/elastic4s/requests/searches/SearchHit.scala
@@ -2,9 +2,9 @@ package com.sksamuel.elastic4s.requests.searches
 
 import com.fasterxml.jackson.annotation.JsonProperty
 import com.sksamuel.elastic4s.Hit
+import com.sksamuel.elastic4s.requests.explain.Explanation
 import com.sksamuel.elastic4s.requests.get.{HitField, MetaDataFields}
 import com.sksamuel.elastic4s.requests.searches.aggs.responses.JacksonSupport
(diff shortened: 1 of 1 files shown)


sksamuel:
👍🏼

PR #2539, opened 2021-10-13
Fix deprecated type

Dbzman:
There's a leftover from the deprecation of the doc type. Since we use that class, we always get that warning.


Dbzman:
It's not that critical and we can already use it since you publish snapshots. Thanks for the offer, though. :)

Issue #909, opened 2017-05-11
Ability to extract whole search response as string/json


halvorgb:
We mainly use elasticsearch for aggregating data, as such all we're interested in when looking at a `SearchResponse` are the aggregations.

An example of a query we run: 
```scala
search(organizationData2Conf.index / "ticket")
      .query { // snip }
      .aggs {
        termsAggregation("by_event_id").size(512).field("event_id")
          .subAggregations(
            revenueAggregation,
            eventStatsPerPeriod(
              period = Month,
              interval = DateHistogramInterval.DAY
            )
          )
      }
      .size(0)
```

We find that the aggregation helper methods are not enough for us, we wind up with `Buckets` from the term aggregations with nowhere to continue.

Could a helper method be added to searchResponse to simply get the whole response as a string or the jackson JsonNode that is used internally? 

We've made some workarounds to get around this issue:

First when using `elastic4s-tcp` we used the `XContent` to build json: https://gist.github.com/halvorgb/296a8b3b7e2591aaeda60e8f6576db48

Now when we're porting to `elastic4s-http` we can't see another way than creating a custom `HttpExecutable[SearchDefinition, Json]` (which is not a pretty solution to say the least): https://gist.github.com/halvorgb/76816db1919925c24916615b659020ea

sksamuel:
Decent request. I'll add this for 5.4.3.
As another issue, what could be added to aggregations so that you don't have to use the raw string.

halvorgb:
Great,  thanks for the swift response.

As for your question: we see no way of getting to sub aggregation results. If the `Bucket` case class had a way of getting sub aggregations it'd be possible to use. We are comfortable defining a schema and validating the responses against it though.

given this query, which has multiple aggregations within a term aggregation within a term aggregation:
```json
{
  "_source": "*",
  "query": {
    "bool": {
      "filter": {
        "term": {
          "is_revoked": false
        }
      }
    }
  },
  "aggs": {
    "by_ticket_type_id": {
      "terms": {
        "field": "ticket_type_id"
      },
      "aggs": {
        "by_checkout_method": {
          "terms": {
            "field": "checkout_method"
          },
          "aggs": {
            "revenue": {
              "sum": {
                "field": "price"
              }
            },
            "total_fee": {
              "sum": {
                "field": "fee"
              }
            }
          }
        }
      }
    }
  },
  "size": 0
}
```

`termsAgg("by_ticket_type_id")` would yield:
```scala
[...]

PR #2559, opened 2021-11-29
Uncomment code for `aggregationsAsString`.

JarrodBaker:
The class `SearchResponse` was moved from 'elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/searches'
to the current location by commit 7f23247f927afa2e884b4ce6ead10bb5c10d86ce.
The only change to this class during this refactor was to comment out
two lines which made available returning the `Aggregations` as a
`String`.

This feature was added as part of sksamuel/elastic4s#909 and is still
useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -646,15 +646,20 @@ Check for the latest released versions on [maven central](http://search.maven.or
 
 ## Building and Testing
 
-This project is built with SBT. So to build
+This project is built with SBT. So to build with:
+
 ```scala
 sbt compile
 ```
 
-And to test
+And to test:
+
 ```scala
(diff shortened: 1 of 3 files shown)


sksamuel:
Looks like some compile errors on 2.13

JarrodBaker:
Sorry about that: I neglected to test against each Scala version. I've updated the code and run both `compile` and `tests` against the cross-build versions and everything appears to be working.

isplendid:
收到了。。。

sksamuel:
👍🏼

PR #2560, opened 2021-12-01
Add support for calendarInterval and fixedInterval to DateHistogramValueSource

destrim:
Changes made due to deprecated warns in project logs from Elasticsearch - field interval is deprecated in Elasticsearch and should not be used.

Added support for `calendarInterval` and `fixedInterval` to `DateHistogramValueSource` case class. 
Added handling of these parameters to `CompositeAggregationBuilder`.

Due to change made in `DateHistogramValueSource` case class - interval is now an Option, just as `calendarInterval` and `fixedInterval` - there is no backward compatibility, `interval` is deprecated and will be removed from Elasticsearch in the future.


sksamuel:
Great thank you !

PR #2570, opened 2021-12-09
Use unified Scaladex badge on ReadMe to show which Scala versions are supported

rtyley:
This unified Scaladex badge summarises which versions of Scala are supported by `elastic4s`, and what the latest `elastic4s` version is for each of those Scala versions, so you don't need to add a new badge every time a new version of Scala is supported:

[elastic4s-core Scala version support](https://index.scala-lang.org/sksamuel/elastic4s/elastic4s-core)

In-situ in the ReadMe file:

image

More details on the badge format: https://github.com/scalacenter/scaladex/pull/660

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,7 @@ elastic4s - Elasticsearch Scala Client
 =========
 
 ![master](https://github.com/sksamuel/elastic4s/workflows/master/badge.svg)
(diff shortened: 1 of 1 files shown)


sksamuel:
That's neat, thanks!

PR #2579, opened 2021-12-13
IndexMappings to ElasticField

### LINKED ISSUE #2548: Howto use IndexMappings as MappingDefinition
I have the following use case. We put documents in ES that have partly a "fixed" structure (i mean some fixed keys for which we defined the "type" and how they are indexed/mapped). Next to that we also want to index other keys of the documents so that we also can search on those.

Now sometimes we need to change the mapping for the fixed keys (typically add some subfield to get other indexing). But since ES does not allow that, we need to create a new index and reindex the old index to the new index. 
However when creating the new index, i want/need to take over the generated mapping of the non-fixed keys of the old index  (to make sure the reindexing works for all documents). To do that, I fetch the mapping of the old index using the `getMapping` request. But that returns the the existing (field) mapping as a `Map[String, Any]`. In the `CreateIndexRequest`, the mappings need to be provided in the `MappingDefinition` as a `Seq[ElasticField]`.

Is there anything already there to transform a `Map[String, Any]` into an ElasticField ? Or any idea how to do the above ?
(I'm actually migrating from elastic4s 7.11.4 to 7.15.1 and in 7.11.4 the MappingDefinition contained a list of FieldDefintions which was a trait, which together with the BasicField, I could in a kind of builder to fold the map to a "correct" FieldDefintion.)

brunoballekens:
proposal to fix #2548 

For compatibility I did not change the `IndexMappings` (i.e. the response objects of the GetMappingRequest).
I've added a "utility" `com.sksamuel.elastic4s.handlers.index.mapping.IndexMappersToFields` that transforms the mappings from an IndexMapping to a `Seq[ElasticField]`.
Since that is the "inverse" function of mapping the ElasticField to json (from the PutMapping req), I added a `construct` method to the ElasticFieldBuilderFn (and all "specialized" FieldBuilderFn) in the hope that if ElasticFields are added, the inverse function is also implemented.
I added also a few unit tests, but I wonder iff and how an exhaustive test suite could be "generated" ?

### DIFF
--- a/elastic4s-domain/src/main/scala/com/sksamuel/elastic4s/fields/AggregateMetricField.scala
+++ b/elastic4s-domain/src/main/scala/com/sksamuel/elastic4s/fields/AggregateMetricField.scala
@@ -1,5 +1,8 @@
 package com.sksamuel.elastic4s.fields
 
+object AggregateMetricField {
+  val `type` = "aggregate_metric_double"
+}
 case class AggregateMetricField(name: String, metrics: Seq[String], defaultMetric: String) extends ElasticField {
-  override def `type`: String = "aggregate_metric_double"
+  override def `type`: String = AggregateMetricField.`type`
 }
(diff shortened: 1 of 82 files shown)


sksamuel:
I like the idea, this will work nicely.

PR #2597, opened 2022-01-05
Update IndexTemplateHandlers.scala

sentenza:
Remove noisy println() from IndexTemplateHandlers


sksamuel:
Nice catch

sksamuel:
Yep, I'll release now.

Issue #2613, opened 2022-02-04
Add missing parameters in word_delimiter_graph in TokenFilter


rtancman:
In `com.sksamuel.elastic4s.analyzers.TokenFilter` miss some parameters for `word_delimiter_graph`. See bellow the undefined list:

- `protected_words`
- `protected_words_path`
- `type_table`
- `type_table_path`

Can be useful?

ref:
- https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-word-delimiter-graph-tokenfilter.html

sksamuel:
Definitely, we should add that. Would you like to raise a PR ?

PR #2620, opened 2022-02-12
Add some parameters in word delimiter token filter

rtancman:
Fix: https://github.com/sksamuel/elastic4s/issues/2613


sksamuel:
Thanks!

Issue #2634, opened 2022-03-07
Add missing parameters in QueryStringQuery


rtancman:
In `com.sksamuel.elastic4s.requests.searches.queries` miss some parameters for `QueryStringQuery`. See bellow the undefined list:

- `quote_analyzer`
- `max_determinized_states`

Can be useful?

Refer:
- https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html

PR #2635, opened 2022-03-07
Add parameters in QueryStringQuery

rtancman:
fix: https://github.com/sksamuel/elastic4s/issues/2634


sksamuel:
And again thanks.

Issue #2625, opened 2022-02-22
Add missing parameters in AnnotatedTextField


rtancman:
In `com.sksamuel.elastic4s.fields` miss some parameters for `AnnotatedTextField `. See bellow the undefined list:

- `copy_to`

Can be useful?

Refer:
- https://www.elastic.co/guide/en/elasticsearch/plugins/current/mapper-annotated-text-usage.html
- https://www.elastic.co/guide/en/elasticsearch/reference/current/copy-to.html#copy-to

sksamuel:
Definitely worth adding

PR #2636, opened 2022-03-08
Add parameter in AnnotatedTextField

rtancman:
fix: https://github.com/sksamuel/elastic4s/issues/2625


sksamuel:
Great thank you!

Issue #2648, opened 2022-04-01
Update Elastic to 8.0.0


rtancman:
Add support for elastic 8.

**Refer:**
- https://www.elastic.co/guide/en/elasticsearch/reference/current/release-notes-8.0.0.html
- https://www.elastic.co/guide/en/elasticsearch/reference/current/migrating-8.0.html#breaking-changes-8.0

PR #2649, opened 2022-04-01
Add support for elastic 8

rtancman:
Fix: https://github.com/sksamuel/elastic4s/issues/2648


elastic4s-core/src/main/scala/com/sksamuel/elastic4s/api/PipelineAggregationApi.scala
```diff
@@ -42,8 +42,8 @@ trait PipelineAggregationApi {
   def minBucketAggregation(name: String, bucketsPath: String): MinBucketPipelineAgg =
     MinBucketPipelineAgg(name, bucketsPath)
 
-  def movingAverageAggregation(name: String, bucketsPath: String): MovAvgPipelineAgg =
-    MovAvgPipelineAgg(name, bucketsPath)
+  def movingFunctionAggregation(name: String, bucketsPath: String, script: String, window: Integer): MovFnPipelineAgg =
```
sksamuel, line 45: was moving average removed ?

rtancman: The moving_avg aggregation has been removed. See the [breaking changes here](https://www.elastic.co/guide/en/elasticsearch/reference/current/migrating-8.0.html#breaking_80_rest_api_changes).

PR #2673, opened 2022-05-30
Fix the charset of the argument byte array passed to the ByteArrayEntity

kenichiro22:
If "file.encoding" property is set, request body of index requests is sent to elasticsearch as the charset specified  by it.

As a result, index requests fail with error like "json_parse_exception,Invalid UTF-8 start byte 0xXX at Source: ByteArrayInputStream" when file.encoding is not utf8.

To fix this problem, fix the charset of the argument byte array passed to the constructor of ByteArrayEntity to utf8.


sksamuel:
👍🏼

PR #2676, opened 2022-06-01
Remove order field of CreateIndexTemplateRequest.

kenichiro22:
There is no order field in current Create index template API.
https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-put-template.html

So the request with order field fails with "unknown field [order]"  error.

Please let me know if it is better to deprecate the field than to delete.


sksamuel:
👍🏼

PR #2680, opened 2022-06-14
Readd custom queries

jackpf:


sksamuel:
Thanks, and yes that's correct.

Issue #2678, opened 2022-06-06
Add support for nested JSON in `_meta`


rney-lookout:
The current implementation of 'Mapping' in IndexHandlers.scala only support single level JSON. Have meta setup such as
```
      "_meta": {
        "presto": {
          "testDoc": { 
            "isArray": true
          }
        }
      }
```
will not parse and the `getIndex` request will return an error

sksamuel:
Yep, bug. Can you open a PR and I'll release.

PR #2682, opened 2022-07-05
Support arbitrary JSON within GetIndex _meta.

hudclark:
Fixes https://github.com/sksamuel/elastic4s/issues/2678

The previous implementation of the GetIndex handler typed the `_meta` field as a Map[String, String]. This causes failures when fetchingindices that use nested JSON within` _meta`.


sksamuel:
👍🏼

PR #2696, opened 2022-07-18
move size to source in reindex request

via119:
Renamed the size parameter to maxDocs as it was done in elastic4s 8.x.
It limits the number of processed documents.
The new size parameter limits the number of documents to reindex per batch.


sksamuel:
👍🏼

PR #2700, opened 2022-08-04
not possible to set script for ExtendedStatsAggregation

shmishleniy:
add script field for extended_stats


sksamuel:
Looks like the tests need a tweak @shmishleniy

PR #2702, opened 2022-08-10
zio 2.0 and zio-json, renamed modules for zio 1.0 and zio-json

justcoon:
zio 2.0 and zio-json 

renamed modules for zio 1.0 and zio-json - added -1 suffix (applied similar pattern like for cats-2, and cats, where latest version is without suffix)


sksamuel:
This is awesome.

Issue #2701, opened 2022-08-10
Bug: PutPipelineRequestHandler missing a leading slash in the URI


grcanosa:
I believe this line is missing a leading slash in the URI making the request fail. 

https://github.com/sksamuel/elastic4s/blob/72aaa55a4a5b0fde70ad711b880cf3da31d03d09/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/ingest/IngestHandlers.scala#L68

Do you want me to create a PR?

sksamuel:
Yes PR would be awesome.

PR #2703, opened 2022-08-11
Added slash to put pipeline uri

grcanosa:
Fixed issue #2701


grcanosa:
You are completely right. I have fixed the tests and run those related with the change and they now run ok. 
Sorry about that!

PR #2731, opened 2022-09-16
Support asynchronous update by query

joelvim:
Current implementation of update by query allows to put `waitForCompletion`to false, however, the response format changes when this field value is false (https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update-by-query.html#docs-update-by-query-task-api). The response will contain only a task id. The current response format does not allow to get the task id.

This PR provides a new API allowing to explicitly make an asynchronous update by query that returns a `GetTask`object allowing to then query for the task progress.

It is meant to be API compatible, type safe and less ambiguous. Current users would not need to modify their working code, that means that code using `UpdateByQueryRequest.waitForCompletion(true)`will remain valid, but it is now deprecated in favour of `updateByQueryAsync`.


sksamuel:
Great change.

sksamuel:
Released 8.4.2

joelvim:
Thank you!

PR #2736, opened 2022-09-29
Scala 3

hughsimpson:
Cross publishes all but 4 modules for scala 3

Also bumps the sttp version significantly... Could be worth leaving in a module for the older version to compile for only scala 2.x.x versions, but the API is entirely unchanged, so might also _not_ be worth bothering with that TBH. It's just one file and honestly I don't see the point in maintaining that version.. may be the wrong approach for this lib though.

Also uglifies the build.sbt to publish only the supported modules for scala 3, which isn't great but at least it keeps everything together


hughsimpson:
@sksamuel I'd previously considered Akka a blocker, but with the upcoming licensing change it's a blocker in its own right anyway, so we've moved off it and discovered the joys of http4s. This is not my favourite pr of all time because of the .string,/.starting() thing  - there's so much inconsistency in the existing codebase that there's no obvious preferred choice... Happy to change those choices if you have a different preference

julienrf:
We would be very interested in using elastic4s from Scala 3. Is there anything we could do to help this PR be merged?

sksamuel:
> @sksamuel I'd previously considered Akka a blocker, but with the upcoming licensing change it's a blocker in its own right anyway, so we've moved off it and discovered the joys of http4s. This is not my favourite pr of all time because of the .string,/.string() thing - there's so much inconsistency in the existing codebase that there's no obvious preferred choice... Happy to change those choices if you have a different preference

Sorry just getting to this.
I would be fine with the akka module being removed - based on licensing changes.
People can still copy n paste the source in if they need it.

We would need to fix the conflicts then I would merge.

hughsimpson:
Cheers. I've merged; let me know if you want any other changes made before this goes in

sksamuel:
This is awesome work. Lets try a release.

julienrf:
Thank you so much!

PR #2746, opened 2022-10-19
Fix AliasField bug of setting the path to the ascii code of last char in the path

cakper:
The existing code results in json: "path": 123 instead of "path": "fieldName" as the field.path is a string and doing foreach on it iterates through characters and sets them one by on the builder


sksamuel:
Thanks!

Issue #2728, opened 2022-09-15
Support for custom aggregations


joelvim:
We have a custom plugin defining our own aggregation. Currently we are using a modified version of elastic4s 5.6.x where we implemented the support for our custom aggregation like any other elastic standard aggregation.

In the process of migrating to Elasticsearch 8, we are trying to remain on a standard version of elastic4s, but we did not find a way to support custom aggregations. The aggregation builder function is a static pattern match that will fail in case an aggregation is not known. That prevents support of custom aggregations.

We were thinking on adding this support. An idea we had was to introduce a parameter in the `AggregationBuilderFn.apply` of type `PartialFunction[AbstractAggregation, XContentBuilder]` or similar and this parameter would bubble up to `SearchHandler` that would become a class taking the partial function as (implicit?) parameter, and a default instance would take the default current behaviour that throws a `NotImplementedError`.

What do you think about something like this? We could do a PR and iterate over it.

sksamuel:
You would need to break the type signature, but if the parameter was implicit, and a default implicit was provided, I think it would be ok.
So yeah lets PR it and see.

joelvim:
Ok let's do it, thanks.

PR #2748, opened 2022-10-24
Support custom aggregations

joelvim:
As discussed previously [here](https://github.com/sksamuel/elastic4s/issues/2728), this pull request aims to bring support of custom aggregations to elastic4s.

The main idea is to replace the default `SearchHandler`that is imported through `ElasticDsl._` by one adding the support of the custom aggregation like this : 
```scala
  private val builder: PartialFunction[AbstractAggregation, XContentBuilder] = {
    case agg: MyCustomAggregation => MyCustomAggregationBuilder(agg)
  }

  private implicit val elasticDsl =
    ElasticDsl.withCustomAggregationHandler(builder)

  import elasticDsl._
```

The advantage of this solution is that it remains API compatible and should not require any modification by anyone not using custom aggregation, however it has a few drawbacks : 
- the above code is not the simplest one, with an import of all the fields of the created `ElasticDsl`
- it is not type safe, so it remains easy to forget to create the handler.

However, a type safe solution would mean a huge refactoring probably breaking the API.

Note that I did not write tests because it would need a plugin deployed on the test ES instance to provide custom aggregations.


sksamuel:
Nice :)

joelvim:
Thank you, and thanks a lot for the release 8.4.4!

PR #2788, opened 2022-12-11
Update elasticsearch 8.5.(0|1) to 8.5.3

Philippus:
I think this is more complete than https://github.com/sksamuel/elastic4s/pull/2787. Previously, some updates were missed I guess.


sksamuel:
Thanks

PR #2790, opened 2022-12-16
feat: supports ranking feature linear query and ext parameters

cdmbr:
- adds support for ranking feature [linear queries](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-rank-feature-query.html#rank-feature-query-linear)
- adds support for [ext parameters](https://github.com/elastic/elasticsearch/blob/main/server/src/main/java/org/elasticsearch/search/SearchExtBuilder.java) (commonly used in plugins)
- adds jusbrasil in company names


sksamuel:
Nice job.

Issue #2674, opened 2022-05-30
Feature request: Support multi_terms Aggregation


iamfiras:
Everything is in the title. It would be cool to support the multi_terms aggregation https://www.elastic.co/guide/en/elasticsearch/reference/8.2/search-aggregations-bucket-multi-terms-aggregation.html

sksamuel:
Yep agreed. If you want to open a PR I'll merge and release.

PR #2799, opened 2023-01-09
Feat: add support for MultiTerms Aggregation

gkfabs:
Here are the changes required to implement MultiTerms Aggregation. That's linked to the issue https://github.com/sksamuel/elastic4s/issues/2674


sksamuel:
Thanks!

PR #2805, opened 2023-01-19
Runtime composite fields

Marc-Ducret:
Runtime composite fields are useful to generate multiple fields with a single script. See https://www.elastic.co/guide/en/elasticsearch/reference/current/runtime-examples.html#runtime-examples-grok-composite for more documentation.


sksamuel:
Thanks!

PR #2808, opened 2023-01-24
Fix nanosecond unit in timeout

### LINKED ISSUE #2807: Timeout unit seems invalid
Currently, when setting a timeout as a `Duration` in a request, in converts it to a string with 
```scala
copy(timeout = (timeout.toNanos + "n").some)
```
However, when using it, we get the following response

```
org.elasticsearch.client.ResponseException: POST http://localhost:58186/_bulk?timeout=300000000000n: HTTP/1.1 400 Bad Request
[error]    {"error":{"root_cause":[{"type":"parse_exception","reason":"failed to parse setting [timeout] with value [300000000000n] as a time value: unit is missing or unrecognized"}],"type":"parse_exception","reason":"failed to parse setting [timeout] with value [300000000000n] as a time value: unit is missing or unrecognized"},"status":400}
```
(This example is taken from bulk API, but the problem is the same on multiple API)

The problem comes from the nanoseconds unit which is abreviated to `n` but the expected one is `nanos` according to the documentation: https://www.elastic.co/guide/en/elasticsearch/reference/current/api-conventions.html#time-units

joelvim:
fixes #2807

### DIFF
--- a/elastic4s-domain/src/main/scala/com/sksamuel/elastic4s/requests/bulk/BulkRequest.scala
+++ b/elastic4s-domain/src/main/scala/com/sksamuel/elastic4s/requests/bulk/BulkRequest.scala
@@ -9,7 +9,7 @@ case class BulkRequest(requests: Seq[BulkCompatibleRequest],
                        timeout: Option[String] = None,
                        refresh: Option[RefreshPolicy] = None) {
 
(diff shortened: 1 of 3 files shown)


sksamuel:
Thanks

PR #2821, opened 2023-02-15
Support slice and slices parameters on Reindex API

### LINKED ISSUE #2820: Add slice and slices parameters in Reindex API.
Add `slice` and `slices` parameters in Reindex API.

flavienbert:
fixed #2820

I reused the `case class Slice(id: String, max: Long)`.
Anyway I don't understand why the `id` param is a type `String`. 
It should be a number. I didn't update it to not break the current API.


sksamuel:
lgtm

sorry for delays @joelvim and @flavienbert.
Next releases will go out tonight.

PR #2850, opened 2023-04-19
Add createOnly flag in reindex

joelvim:


joelvim:
There is a test failing, but it is not related to the PR, and passes in local environment. It seems the problem comes from a timeout waiting for the first shard to be active.

sksamuel:
👍🏻

joelvim:
Thank you

sksamuel:
7.10.11 has been released.

PR #2896, opened 2023-08-07
Add failures field to UpdateByQueryResponse and DeleteByQueryResponse

### LINKED ISSUE #2883: DeleteByQueryResponse should have failures field
According to [Elasticsearch's documentation](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/docs-delete-by-query.html#docs-delete-by-query-api-response-body), `DeleteByQueryResponse` is supposed to have `failures` field, but it doesn't have one at the moment.

azolotko:
Also update `ReindexResponse` to make its `failures` field have the same type as the one in `UpdateByQueryResponse` and `DeleteByQueryResponse`.

Resolves #2883

### DIFF
--- a/elastic4s-handlers/src/main/scala/com/sksamuel/elastic4s/handlers/reindex/ReindexHandlers.scala
+++ b/elastic4s-handlers/src/main/scala/com/sksamuel/elastic4s/handlers/reindex/ReindexHandlers.scala
@@ -6,13 +6,11 @@ import com.sksamuel.elastic4s.handlers.ElasticErrorParser
 import com.sksamuel.elastic4s.requests.common.RefreshPolicyHttpValue
 import com.sksamuel.elastic4s.requests.reindex.ReindexRequest
 import com.sksamuel.elastic4s.requests.task.CreateTaskResponse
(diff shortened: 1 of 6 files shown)


elastic4s-handlers/src/main/scala/com/sksamuel/elastic4s/handlers/reindex/ReindexHandlers.scala
```diff
@@ -27,7 +25,7 @@ case class ReindexResponse(took: Long,
                            @JsonProperty("throttled_millis") throttledMillis: Long,
                            @JsonProperty("requests_per_second") requestsPerSecond: Long,
                            @JsonProperty("throttled_until_millis") throttledUntilMillis: Long,
-                           failures: Seq[ReindexFailure]) {
+                           failures: Option[Seq[BulkIndexByScrollFailure]]) {
```
azolotko, line 28: [specification/_global/reindex/ReindexResponse.ts#L31](https://github.com/elastic/elasticsearch-specification/blob/656080dee792f93a849cd7fbf566f528f858a5ea/specification/_global/reindex/ReindexResponse.ts#L31)

sksamuel:
lgtm

PR #2897, opened 2023-08-11
Feat: add support for date format in runtime mappings

gkfabs:
Runtime mappings support format attribute:

https://www.elastic.co/guide/en/elasticsearch/reference/8.9/runtime-mapping-fields.html
Runtime fields with a type of date can accept the [format](https://www.elastic.co/guide/en/elasticsearch/reference/8.9/mapping-date-format.html) parameter exactly as the date field type.

I am not sure if any type checking on field to corroborate it is a DateField is necessary.

Anyhow elastic returns an error if format is not on a DateField.


sksamuel:
lgtm

PR #2899, opened 2023-08-15
Add Pekko support

dragisak:
* elastic4s-client-pekko (including Scala 3)
* elastic4s-streams-pekko (including Scala 3)
* Removed ScalaMock from Akka and Pekko client tests
* Upgrade Scala 3 to Scala 3.3.0

Just a copy of Akka code with some renames.


sksamuel:
Looks good - will merge when S3.3 support added

dragisak:
I also replaced ScalaMock with Mockito so we can compile Pekko client to Scala 3

sksamuel:
awesome :)

PR #2925, opened 2023-09-26
Reduce buildBulkHttpBody's retained set by making BulkBuilderFn return an Iterator

azolotko:
There's no need to keep all the rows in memory at the same time.


sksamuel:
Thanks!

PR #2942, opened 2023-10-25
[#2941] Implement script score query

### LINKED ISSUE #2941: Implement ScriptScoreQuery
# TLDR;

I'd like to use the ElasticSearch [script score query ](https://www.elastic.co/guide/en/elasticsearch/reference/8.10/query-dsl-script-score-query.html). I don't believe it's implemented in `elastic4s`. Would it be acceptable to create a PR that implements it?

# Details

The ElasticSearch Search DSL supports a script score query since v7.0. See docs for [v8.10 (latest)](https://www.elastic.co/guide/en/elasticsearch/reference/8.10/query-dsl-script-score-query.html) and [v7.0](https://www.elastic.co/guide/en/elasticsearch/reference/7.0/query-dsl-script-score-query.html).

The example HTTP request from the docs looks like:

```
GET /_search
{
  "query": {
    "script_score": {
      "query": {
        "match": { "message": "elasticsearch" }
      },
      "script": {
        "source": "doc['my-int'].value / 10 "
      }
    }
  }
}
```

As far as I can see, elastic4s doesn't support constructing such a query. There _does_ exist the [`ScriptScoreQuery` class](https://github.com/sksamuel/elastic4s/blob/master/elastic4s-domain/src/main/scala/com/sksamuel/elastic4s/requests/searches/queries/ScriptScoreQuery.scala) but the writing of this class is not implemented in [`QueryBuilderFn`](https://github.com/sksamuel/elastic4s/blob/bff9d655b04f52cdf7f2c08c72dcee2e51157620/elastic4s-handlers/src/main/scala/com/sksamuel/elastic4s/handlers/searches/queries/QueryBuilderFn.scala#L59).

I found this stackoverflow from a user in 2020 searching for the same functionality: https://stackoverflow.com/questions/65238974/scriptscorequery-implementation-in-elastic4s

sam-briggs-depop:
Implements the script score query to fix #2941.
For now the ScriptScoreQueryBodyFn2 class has the terrible name because
I suspect that the current ScriptScoreQueryBodyFn is never exercised.
The ScriptScore that it accepts isn't actually a query, so will never
hit the pattern match?


elastic4s-tests/src/test/scala/com/sksamuel/elastic4s/search/queries/ScriptScoreQueryTest.scala
```diff
@@ -0,0 +1,57 @@
...
+  "script score query" should {
+    "filter by query" in {
+      client.execute {
+        search("person") postFilter {
+          scriptScoreQuery(termQuery("name", "finch"))
+            .script(Script("doc['age'].value % 3"))
+        }
+      }.await.result.totalHits shouldBe 3
+    }
+    "sort by score" in {
```
sam-briggs-depop, line 48: This test is failing currently. The score appears to be unaffected.

sksamuel:
The PR is still in draft. If it's ready, bar that other test, please move to ready state.

elastic4s-tests/src/test/scala/com/sksamuel/elastic4s/search/queries/ScriptScoreQueryTest.scala
briggySmalls: Found my error. I'd copy/pasted `search(<indexname>) postFilter ...` rather than `search(<indexname>) query ...`

PR #2952, opened 2023-11-09
Add support for PIT API

### LINKED ISSUE #2951: Add support for PIT in search
In order to be able to use search with search_after using a PIT, we need to add the `pit` parameter in the search request and to add the `_pit` creation and deletion API.

joelvim:
Support creating and deleting PIT, and allow search requests using PIT
Fixes #2951


sksamuel:
nice!

PR #2955, opened 2023-11-11
2941/implement script score query

sam-briggs-depop:
Implements the script score query to fix https://github.com/sksamuel/elastic4s/issues/2941.
For now the ScriptScoreQueryBodyFn2 class has the terrible name because
I suspect that the current ScriptScoreQueryBodyFn is never exercised.
The ScriptScore that it accepts isn't actually a query, so will never
hit the pattern match?

Sorry. I closed and reopened because I rebased and got into a pickle


sksamuel:
Nice !

PR #2965, opened 2023-11-29
fix: handle errors in multisearch response

cdmbr:
bug: multisearch response with a non 2XX status is handled as successful response (MultiSearchResponse(Nil))

sample request:

```
GET /_msearch
{}
{"query" : {"exists" : { "field": ""}}}
{}
{"query" : {"exists" : { "field": "ok"}}}
```

```
{
  "error": {
    "root_cause": [
      {
        "type": "illegal_argument_exception",
        "reason": "field name is null or empty"
      }
    ],
    "type": "illegal_argument_exception",
    "reason": "field name is null or empty"
  },
  "status": 400
}
```


cdmbr:
fyi @sksamuel

sksamuel:
lgtm

Issue #2976, opened 2023-12-14
WAND optimization with a numeric track_total_hits parameter


apavlychev:
Hello!
How to set `track_total_hits` equal a numeric value (100) via `elastic4s` in a case WAND optimization.
```
GET my-index-000001/_search
{
  "track_total_hits": 100,
  "query": {
    "match": {
      "user.id": "elkbee"
    }
  }
}
```

P.S. materials
https://www.elastic.co/guide/en/elasticsearch/reference/8.11/search-your-data.html#track-total-hits
https://www.elastic.co/blog/faster-retrieval-of-top-hits-in-elasticsearch-with-block-max-wand

PR #2977, opened 2023-12-18
WAND optimization with a numeric track_total_hits parameter

apavlychev:
close https://github.com/sksamuel/elastic4s/issues/2976


sksamuel:
👍🏻

PR #2992, opened 2024-01-22
fix: autofield iterable serialization

cdmbr:
collections like set, list set were not properly serialized to json while using autofield in xcontentbuilder

e.g.:

```
 XContentFactory.obj().autofield("iterable", Set(1, 2, 3)).string
```
yields:

```
{"iterable":["Set(1, 2, 3)"]}
```

while it should be:

```
{"iterable":[1,2,3]}
```


israelsimas (approved):

sksamuel:
lgtm

PR #2996, opened 2024-01-26
Add `failed_shards` in task error

joelvim:
Same as in `ElasticError`, add the field `failed_shards` in `TaskError` to be able to get more details when a task fails.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -26,7 +26,7 @@ jobs:
           java-version: 11
 
       - name: Launch elastic docker
-        run: docker run -d -it -p 39227:9200 -p 39337:9300 -e "discovery.type=single-node" -v /home/runner/work/elastic4s/elastic4s/elastic4s-tests/src/test/resources/elasticsearch.yml:/usr/share/elasticsearch/config/elasticsearch.yml docker.elastic.co/elasticsearch/elasticsearch:8.11.4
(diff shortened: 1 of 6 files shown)


joelvim:
I don't understand why test is failing on CI while it is ok in local. It seems that the test trying to create a task that fails to check the error is correctly parsed doesn't create a task that fails 😢 I create a query that is too big (>90000 clauses) to be sure it fails, but that doesn't seem to work 😞

joelvim:
It seems that there is a discrepency in the versions of servers used for tests, and it impacts the results of the tests : #3000

joelvim:
Failing tests should be fixed after merging #3001

sksamuel:
Tests still not passing unfortunately.

joelvim:
I don't understand why it fails. In local it passes. I try to create a task that I am sure it will fail, but it seems to try to get the task before ES executes it, so it did not yet have the error. I tried to put a sleep, and then to increase the sleep time but it doesn't change anything. Maybe the CI ES server is too slow?

sksamuel:
Could try extending the time outs or something

joelvim:
I tried to increase up to 1 minute, but it doesn't change anything ☹️

joelvim:
I realised I was wrong. I thought the task did not complete, but actually it completes, but does not fail. Do you see a reason to have such a difference in behaviour? I create a query with a number of clauses exceeding the maximum allowed number, so it should fail anyway. I don't see anywhere a configuration overriding the default number of allowed clauses. Maybe there is a better way to generate an error.

joelvim:
I finally made the test pass by increasing the number of clauses, because the CI machine has more ram than mine, so it accepted more clauses.

joelvim:
Hello @Philippus It seems you're taking over the project. Could you give a look to this PR please? Thanks a lot.

Philippus (approved):
LGTM

PR #2997, opened 2024-01-31
update sbt

yanns:


sksamuel:
The auto update put us to 1.9,9 so will close this

PR #3036, opened 2024-04-05
Fix cancel tasks API

joelvim:
Elasticsearch provides two ways to cancel a task
* by its id : `POST _tasks/oTUltX4IQMOUUVeiohTt8A:12345/_cancel`
* by node & actions : `POST _tasks/_cancel?nodes=nodeId1,nodeId2&actions=*reindex`

Elastic4s which takes nodeIds and actions, provides a single interface for both, but both are broken : 
* if no node id is provided, the endpoint returned is `s"/_tasks/cancel"` while it should be `s"/_tasks/_cancel"`
* if node ids are provided, it returns `s"/_tasks/task_id:${request.nodeIds.mkString(",")}/_cancel"` but this URL is invalid for two reasons:
    * task_id in the path breaks the task id that is provided, 
    * the API does not support more than one task. 
Both result in a 400 with `"reason": "malformed task id task_id:<my_task_id>"`

As the handler always returns a Right `override def handle(response: HttpResponse) = Right(response.statusCode >= 200 && response.statusCode < 300)` this API always fail silently.

My solution is to provide a new API `cancelTaskById(taskId: String)` to prevent ambiguities from the use of the other API that becomes a pure nodeIds + actions API.

### DIFF
--- a/elastic4s-tests/src/test/scala/com/sksamuel/elastic4s/tasks/CancelTaskTest.scala
+++ b/elastic4s-tests/src/test/scala/com/sksamuel/elastic4s/tasks/CancelTaskTest.scala
@@ -35,7 +35,6 @@ class CancelTaskTest extends AnyWordSpec with Matchers with DockerTests {
       val response = client.execute {
         cancelTaskById(resp.nodeId, resp.taskId)
       }.await
-      println(response)
       response.result should be(true)
     }
 
@@ -50,7 +49,6 @@ class CancelTaskTest extends AnyWordSpec with Matchers with DockerTests {
         cancelTasks(resp.nodeId).actions("*reindex")
       }.await
 
-      println(response)
       response.result should be(true)
     }
   }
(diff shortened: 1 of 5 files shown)


elastic4s-tests/src/test/scala/com/sksamuel/elastic4s/tasks/CancelTaskTest.scala
```diff
@@ -0,0 +1,59 @@
...
+      // kick off a task
+      val resp = client.execute {
+        reindex("cancel_task_a", "cancel_task_b").waitForCompletion(false)
+      }.await.result.right.get
+
+      // use the task id from the above task
+      val response = client.execute {
+        cancelTaskById(resp.nodeId, resp.taskId)
+      }.await
+      println(response)
```
Philippus, line 38: is this println needed?

PR #3081, opened 2024-06-06
feat: add value_as_string to avg/min/max/sum aggregations

mcenkar:
min/max aggregations return also `value_as_string`, which can be useful when running over date fields:

```json
{
  "aggregations": {
    "max-example": {
      "value": 1770336000000,
      "value_as_string": "2026-02-06"
    },
    "min-example": {
      "value": 1716422400000,
      "value_as_string": "2024-05-23"
    }
  }
}
```
It is possible to access them now.

image

### DIFF
--- a/elastic4s-domain/src/main/scala/com/sksamuel/elastic4s/requests/searches/aggs/responses/aggresponses.scala
+++ b/elastic4s-domain/src/main/scala/com/sksamuel/elastic4s/requests/searches/aggs/responses/aggresponses.scala
@@ -78,14 +78,14 @@ object SignificantTermsAggResult {
   )
 }
 
-case class AvgAggResult(name: String, valueOpt: Option[Double]) extends MetricAggregation {
+case class AvgAggResult(name: String, valueOpt: Option[Double], valueAsString: Option[String]) extends MetricAggregation {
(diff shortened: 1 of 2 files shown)


mcenkar:
Hi @Philippus,

appreciate if you could have a look at this PR. I can also add to `avg` or `sum` aggs, it makes less sense for me over dates, but others might find it useful.

Philippus:
It would be awesome if you could add them for `avg` and `sum` as well. Would it also be possible to add some tests?

mcenkar:
I updated PR + added some tests.

For `avg` and `sum` they are in separate tests ([first](https://github.com/Philippus/elastic4s/blob/main/elastic4s-tests/src/test/scala/com/sksamuel/elastic4s/requests/searches/aggs/SumAggregationHttpTest.scala), [second](https://github.com/Philippus/elastic4s/blob/main/elastic4s-tests/src/test/scala/com/sksamuel/elastic4s/requests/searches/aggs/AvgAggregationHttpTest.scala)) so more a bit more work but I can add them if you want.

Philippus (approved):

Philippus:
thanks!

PR #3090, opened 2024-06-18
Support Update By Query with slices auto

flavienbert:
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update-by-query.html#docs-update-by-query-slice  According to the doc, we should be able to use the value **auto** with the param **slices**

### DIFF
--- a/elastic4s-domain/src/main/scala/com/sksamuel/elastic4s/requests/common/Slices.scala
+++ b/elastic4s-domain/src/main/scala/com/sksamuel/elastic4s/requests/common/Slices.scala
@@ -1,0 +1,6 @@
+package com.sksamuel.elastic4s.requests.common
+
+sealed trait Slices
+
+case object AutoSlices extends Slices
+case class NumericSlices(slices: Int) extends Slices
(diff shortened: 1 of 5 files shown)


flavienbert:
@Philippus I don't understand the error on the scala-2_12 CI. Can you help me?

Philippus (changes requested):
We cannot change the existing parameters of `UpdateByQueryAsyncRequest` and `UpdateByQueryRequest`, because that would break compatibility.
What you could do is add an extra parameter, maybe `slicesAuto: Option[Boolean]`, default set to None.

In the handler it would then become
```
if (request.slicesAuto.getOrElse(false))
  params.put("slices", "auto")
else
  request.slices.foreach(params.put("slices", _))
```

It would be awesome if you could add the auto slices option to `ReindexRequest` and `DeleteByQueryRequest` as well.

flavienbert:
Actually, it partially breaks compatibility because I assume most of people use the DSL  
`def slices(slices: Int): UpdateByQueryRequest = copy(slices = Some(NumericSlices(slices)))`.
We can also add an implicit conversion Int -> Slices in the scope to avoid breaking changes.

The drawback having 2 different params for the slices is that a user can use both in same time, so which one should we keep? It can be confusing.

What do you think?

Philippus:
We can add a new parameter like `numericOrAutoSlices: Option[Slices] = None` and deprecate the old parameter. The new parameter should take precedence over the deprecated one. The methods then can look like this:
```
  def slices(slices: Int): ReindexRequest = copy(numericOrAutoSlices = NumericSlices(slices).some)
  def slicesAuto(): ReindexRequest = copy(numericOrAutoSlices = AutoSlices.some)
```
and in the handlers we could do something like this:
```
      request.numericOrAutoSlices.foreach {
        case AutoSlices            => params.put("slices", "auto")
        case NumericSlices(slices) => params.put("slices", slices.toString)
      }
      if (request.numericOrAutoSlices.isEmpty)
        request.slices.foreach(params.put("slices", _))
```

PR #3099, opened 2024-06-27
DenseVectorField lacks index_options parameter

apavlychev:
close https://github.com/Philippus/elastic4s/issues/3098

https://www.elastic.co/guide/en/elasticsearch/reference/current/dense-vector.html#dense-vector-params

### DIFF
--- a/elastic4s-handlers/src/main/scala/com/sksamuel/elastic4s/handlers/fields/DenseVectorFieldBuilderFn.scala
+++ b/elastic4s-handlers/src/main/scala/com/sksamuel/elastic4s/handlers/fields/DenseVectorFieldBuilderFn.scala
@@ -20,7 +20,7 @@ object DenseVectorFieldBuilderFn {
   def toField(name: String, values: Map[String, Any]): DenseVectorField = DenseVectorField(
     name,
     values.get("dims").map(_.asInstanceOf[Int]).get,
-    indexOptions = values.get("index_options").map(_.asInstanceOf[Map[String, Any]]).map(getIndexOptions),
+    indexOptions = values.get("index_options").map(_.asInstanceOf[Map[String, Any]]).map(getIndexOptions)
   )
 
 
--- a/elastic4s-handlers/src/main/scala/com/sksamuel/elastic4s/handlers/fields/DenseVectorFieldBuilderFn.scala
+++ b/elastic4s-handlers/src/main/scala/com/sksamuel/elastic4s/handlers/fields/DenseVectorFieldBuilderFn.scala
@@ -1,28 +1,31 @@
 package com.sksamuel.elastic4s.handlers.fields
 
-import com.sksamuel.elastic4s.fields.{DenseVectorField, DenseVectorIndexOptions, Flat, Hnsw, Int8Flat, Int8Hnsw}
+import com.sksamuel.elastic4s.fields.{DenseVectorField, DenseVectorIndexOptions, FlatIndexOptions, HnswIndexOptions, Int8FlatIndexOptions, Int8HnswIndexOptions}
 import com.sksamuel.elastic4s.json.{XContentBuilder, XContentFactory}
 
 object DenseVectorFieldBuilderFn {
 
-  private def getIndexOptions(values: Map[String, Any]) = DenseVectorIndexOptions(
+  private def getIndexOptions(values: Map[String, Any]): DenseVectorIndexOptions=
     values("type").asInstanceOf[String] match {
-      case "hnsw" => Hnsw
-      case "int8_hnsw" => Int8Hnsw
-      case "flat" => Flat
-      case "int8_flat" => Int8Flat
-    },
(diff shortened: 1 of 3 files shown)


elastic4s-handlers/src/main/scala/com/sksamuel/elastic4s/handlers/fields/DenseVectorFieldBuilderFn.scala
```diff
@@ -1,12 +1,26 @@
...
+    values.get("m").map(_.asInstanceOf[Int]),
+    values.get("ef_construction").map(_.asInstanceOf[Int]),
+    values.get("confidence_interval").map(_.asInstanceOf[Double])
+  )
+
   def toField(name: String, values: Map[String, Any]): DenseVectorField = DenseVectorField(
     name,
-    values.get("dims").map(_.asInstanceOf[Int]).get
+    values.get("dims").map(_.asInstanceOf[Int]).get,
+    indexOptions = values.get("index_options").map(_.asInstanceOf[Map[String, Any]]).map(getIndexOptions),
```
Philippus, line 23:
```suggestion
    indexOptions = values.get("index_options").map(_.asInstanceOf[Map[String, Any]]).map(getIndexOptions)
```

elastic4s-handlers/src/main/scala/com/sksamuel/elastic4s/handlers/fields/DenseVectorFieldBuilderFn.scala
```diff
@@ -17,6 +31,14 @@ object DenseVectorFieldBuilderFn {
     builder.field("dims", field.dims)
     builder.field("index", field.index)
     builder.field("similarity", field.similarity.name)
+    field.indexOptions.foreach { options =>
+      builder.startObject("index_options")
+      builder.field("type", options.`type`.name)
+      options.m.foreach(builder.field("m", _))
```
Philippus, line 37: would be nice if we add `m` only for `Hnsw` and `Int8Hnsw`.
Something like `if (Seq(Hnsw, Int8Hnsw).contains(options.`type`)) options.m.foreach(builder.field("m", _))`

elastic4s-handlers/src/main/scala/com/sksamuel/elastic4s/handlers/fields/DenseVectorFieldBuilderFn.scala
```diff
@@ -17,6 +31,14 @@ object DenseVectorFieldBuilderFn {
...
+      options.efConstruction.foreach(builder.field("ef_construction", _))
```
Philippus, line 38: would be nice if we add efConstruction only for `Hnsw` and `Int8Hnsw`.

elastic4s-handlers/src/main/scala/com/sksamuel/elastic4s/handlers/fields/DenseVectorFieldBuilderFn.scala
```diff
@@ -17,6 +31,14 @@ object DenseVectorFieldBuilderFn {
     builder.field("dims", field.dims)
     builder.field("index", field.index)
     builder.field("similarity", field.similarity.name)
+    field.indexOptions.foreach { options =>
```
Philippus, line 34: I think we need to also check if `index` is `true` here, because "This parameter can only be specified when index is true." according to the docs.

Philippus (changes requested):
Can you add some tests in `ElasticFieldBuilderFnTest` and `DenseVectorFieldTest` ?

elastic4s-handlers/src/main/scala/com/sksamuel/elastic4s/handlers/fields/DenseVectorFieldBuilderFn.scala
apavlychev: refactor ADT for a `DenseVectorIndexOptions`

elastic4s-handlers/src/main/scala/com/sksamuel/elastic4s/handlers/fields/DenseVectorFieldBuilderFn.scala
apavlychev: refactor ADT for a `DenseVectorIndexOptions`

elastic4s-handlers/src/main/scala/com/sksamuel/elastic4s/handlers/fields/DenseVectorFieldBuilderFn.scala
apavlychev: add a filter and tests

apavlychev:
@Philippus what do you think about this changes ?

Philippus:
Thanks! I will work on DenseVectorField a bit more outside of this PR before creating a new release.

PR #3101, opened 2024-06-28
Knn query lacks inner_hits field

apavlychev:
close https://github.com/Philippus/elastic4s/issues/3100

https://www.elastic.co/guide/en/elasticsearch/reference/current/knn-search.html#nested-knn-search
https://www.elastic.co/guide/en/elasticsearch/reference/current/inner-hits.html
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-fields.html#search-fields-param


Philippus (changes requested):
Could you add a test, showing how it works in practice?

apavlychev:
> Could you add a test, showing how it works in practice?

added `ChunkKnnTest` tests with examples from https://www.elastic.co/guide/en/elasticsearch/reference/current/knn-search.html#nested-knn-search

PR #3113, opened 2024-07-13
Improve DenseVectorField

Philippus:
Applying this PR will improve DenseVectorField, among which:
- add element_type parameter
- add max_inner_product similarity metric
- introduce new apply method with proper defaults (following the documentation)
- not set similarity if index is set to false


TactiC (approved):
LGTM

PR #3118, opened 2024-07-16
Add client for http4s library

igor-vovk:
Hi! I don't know if there is space for another client, but it seemed appropriate for me to add it, since it's kinda mainstream client in `cats-effect` ecosystem (http4s/circe/cats-effect). I was kinda surprised not to find it, ended up implementing it, but got tired copying it between my projects where I use elastic4s.

If you think it's a good idea to add it, I will be happy to add tests in this PR as well. Thank you.


igor-vovk:
It just came to me, @Philippus if you think that `elastic4s` supports enough clients out of the box (I found that sttp for example already has http4s backend), I can just publish it as a separate library.

lenguyenthanh:
I would love to use this client! I think currently sttp use future backend (which is backed scala future & akka iiuc). So this is definitely worth adding imho.

Also We probably can have more perform IO backed client if We have this: https://github.com/Philippus/elastic4s/issues/2379

igor-vovk:
@lenguyenthanh I was also thinking about that, addressed in https://github.com/Philippus/elastic4s/pull/3119

lenguyenthanh:
> @lenguyenthanh I was also thinking about that, addressed in #3119

this is great, thanks! I was about to do it.

Philippus:
I think adding a new client is fine, provided someone is using it in production (like you). Maybe we can mark it as experimental, allowing us to make backwards incompatible changes for a certain time.

igor-vovk:
@Philippus nice! Then I'm going to add tests for it

igor-vovk:
@Philippus I've added some tests following the pattern from sttp. I think it is ready to be merged, WDYT?

Philippus:
@igor-vovk now that https://github.com/Philippus/elastic4s/pull/3124 is merged, care to get back and clean up this PR (or create a new one based on the current main branch)?

Philippus:
@lenguyenthanh Do you have time to take a look at the PR to see if this will work for you as well?

lenguyenthanh:
> @lenguyenthanh Do you have time to take a look at the PR to see if this will work for you as well?

of course, I'll take a look tomorrow.

elastic4s-client-http4s/src/main/scala/com/sksamuel/elastic4s/http4s/Elastic4sEntityEncoders.scala
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+  implicit def optionalEntityEncoder[F[_], A](implicit ee: http4s.EntityEncoder[F, A]): http4s.EntityEncoder[F, Option[A]] =
+    new http4s.EntityEncoder[F, Option[A]] {
+      override def toEntity(a: Option[A]): http4s.Entity[F] = {
+        a match {
+          case Some(value) => ee.toEntity(value)
+          case None => http4s.Entity.empty
+        }
+      }
```
lenguyenthanh, line 38:
```suggestion
    new http4s.EntityEncoder[F, Option[A]] {
      override def toEntity(a: Option[A]): http4s.Entity[F] =
        a.fold[http4s.Entity[F]](http4s.Entity.empty)(ee.toEntity)
```
nit

elastic4s-client-http4s/src/main/scala/com/sksamuel/elastic4s/http4s/RequestResponseConverters.scala
```diff
@@ -0,0 +1,52 @@
...
+import scala.language.higherKinds
+
+trait RequestResponseConverters extends Elastic4sEntityEncoders {
+
+  def elasticRequestToHttp4sRequest[F[_] : Async : Files](
+    endpoint: ElasticNodeEndpoint,
+    request: elastic4s.ElasticRequest,
+  ): http4s.Request[F] = {
+    val uri = http4s.Uri(
+      scheme = Some(http4s.Uri.Scheme.unsafeFromString(endpoint.protocol)),
```
lenguyenthanh, line 19:
```suggestion
      scheme = http4s.Uri.Scheme.fromString(endpoint.protocol).toOption,
```
nit

elastic4s-client-http4s/src/main/scala/com/sksamuel/elastic4s/http4s/RequestResponseConverters.scala
```diff
@@ -0,0 +1,52 @@
...
+      authority = Some(http4s.Uri.Authority(
```
lenguyenthanh, line 20: as we already have `cats.syntax.all._` we can use `.some` as it nicer imho.

lenguyenthanh (approved):
I only have few nit comments, otherwise lgtm. And this will be so much better with the other of @igor-vovk's pr: https://github.com/Philippus/elastic4s/pull/3119

igor-vovk:
@lenguyenthanh I've addressed all your comments

elastic4s-client-http4s/src/test/scala/com/sksamuel/elastic4s/http4s/Http4sRequestHttpClientTest.scala
```diff
@@ -0,0 +1,35 @@
...
+  private val http4sClient = Http4sClient.usingIO(
+    http4s,
+    ElasticNodeEndpoint("http", elasticHost, elasticPort.toInt, None),
+  )
+  override val client: ElasticClient = ElasticClient(http4sClient)
+
+  "Http4sRequestHttpClient" should "be able to call elasticsearch" in {
+    client.execute {
+      catHealth()
+    }.await.result.status shouldBe "green"
```
Philippus, line 22: Looks like some times the test gives back `yellow`, which I think for this test is ok.

Philippus (changes requested):

elastic4s-client-http4s/src/test/scala/com/sksamuel/elastic4s/http4s/Http4sRequestHttpClientTest.scala
igor-vovk: @Philippus fixed. I think it would be a good idea to extract a common test suite for all http clients (since they all can be tested in a single way), will try to do it in a separate PR, and then more tests can be added

igor-vovk:
@Philippus I also noticed test flakiness, tried to add health check before proceeding on the level of GH actions in #3207, but failing to understand why it fails in GH Actions environment (works locally, probably something with bash options...)

Philippus:
I think it's because of all the tests running on the same instance. If you change the test to "yellow" or "green" I think it will be fine.

igor-vovk:
@Philippus I think what happens is that this status changes from `red` to `yellow` and then to `green` (doc https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-health.html). And then depending on when the test runs it can receive any of these, thus flakiness of the test. In any case, I've deleted this test. Should I bring it back?

igor-vovk:
Marked #3207 as ready for review. It now waits elastic to start before continuing, should reduce tests flakiness

Philippus:
> @Philippus I think what happens is that this status changes from `red` to `yellow` and then to `green` (doc https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-health.html). And then depending on when the test runs it can receive any of these, thus flakiness of the test. In any case, I've deleted this test. Should I bring it back?

If you think it's worth having it as a test I think we can bring it back. We can always remove it again, if it turns out to be flaky again.

igor-vovk:
Reverted the deletion of the test. After #3207 will be merged, the test should stop being flaky

PR #3124, opened 2024-08-08
Support authentication on the level of ElasticClient

igor-vovk:
Addresses https://github.com/Philippus/elastic4s/issues/3122. I copied authentication settings that were introduced in http4s PR https://github.com/Philippus/elastic4s/pull/3118, will be happy to remove them from http4s PR after this will be merged.
 
Also, they are now can be removed from AkkaHttpClient, that also implements adding credentials by itself.

By defining credentials on the ElasticClient level there is now a single place where credentials are used, so there is no need to rely on client-specific implementations of credential passing.


elastic4s-core/src/main/scala/com/sksamuel/elastic4s/ElasticClient.scala
```diff
@@ -60,12 +63,49 @@ case class ElasticClient(client: HttpClient) extends AutoCloseable {
...
+  case class ApiKey(apiKey: String) extends Authentication
+
+  case object NoAuth extends Authentication
+}
+
+case class CommonRequestOptions(
+  timeout: Duration,
+  masterNodeTimeout: Duration,
+  authentication: Authentication,
+  headers: Map[String, String] = Map.empty
```
Philippus, line 101: Can you switch these around, for slightly better backwards compatibility?

Philippus (changes requested):
Would it be possible to add some tests for the different cases?

elastic4s-core/src/main/scala/com/sksamuel/elastic4s/ElasticClient.scala
igor-vovk: Fixed, and gave it default value

Philippus (approved):

igor-vovk:
Sorry @Philippus I haven't added tests yet. I was thinking whether I should make an IT-test, e.g. try to setup a credentials in ES, and then test against it, or just do a unit test? I was kind of blocked by not being able to choose, maybe you could share what would be the preferred way?

Philippus:
> Sorry @Philippus I haven't added tests yet. I was thinking whether I should make an IT-test, e.g. try to setup a credentials in ES, and then test against it, or just do a unit test? I was kind of blocked by not being able to choose, maybe you could share what would be the preferred way? Will be happy to add tests after ☺️

I think ideally, set up an Elasticsearch test container requiring credentials and test against that. But I don't know how easy it would be to set that up. I'm fine with someone adding that in a later PR.

PR #3125, opened 2024-08-09
Fix missing slashes

ddvleeuwen:
In a previous PR we fixed 400 errors by adding a missing `/`. Point is, we forgot to add them for all the other _synonyms endpoints. https://github.com/Philippus/elastic4s/pull/3116


TactiC (approved):

Philippus (approved):
Oops, LGTM.

PR #3147, opened 2024-08-29
Support automatic slicing using constants

Philippus:
for UpdateByQuery, Reindex and DeleteByQuery. This way we don't have to change the signature of the case class at all. I took the inspiration from how it works in Elasticsearch itself (https://github.com/elastic/elasticsearch/blob/main/server/src/main/java/org/elasticsearch/index/reindex/AbstractBulkByScrollRequest.java#L39).

### DIFF
--- a/elastic4s-handlers/src/main/scala/com/sksamuel/elastic4s/handlers/update/UpdateHandlers.scala
+++ b/elastic4s-handlers/src/main/scala/com/sksamuel/elastic4s/handlers/update/UpdateHandlers.scala
@@ -6,7 +6,7 @@ import com.sksamuel.elastic4s.handlers.common.FetchSourceContextQueryParameterFn
 import com.sksamuel.elastic4s.handlers.script.ScriptBuilderFn
 import com.sksamuel.elastic4s.handlers.searches.queries
 import com.sksamuel.elastic4s.json.{XContentBuilder, XContentFactory}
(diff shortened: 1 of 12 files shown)


elastic4s-handlers/src/main/scala/com/sksamuel/elastic4s/handlers/update/UpdateHandlers.scala
```diff
@@ -87,10 +87,9 @@ trait UpdateHandlers {
       request.scrollSize.foreach(params.put("scroll_size", _))
       request.waitForActiveShards.foreach(params.put("wait_for_active_shards", _))
       request.waitForCompletion.foreach(params.put("wait_for_completion", _))
-      request.slices.foreach {
-        case AutoSlices            => params.put("slices", "auto")
-        case NumericSlices(slices) => params.put("slices", slices)
-      }
+      request.slices.foreach(s =>
+        if (s == Slicing.AutoSlices) params.put("slices", Slicing.AutoSlicesValue) else params.put("slices", s)
```
TactiC, line 91: should this be 
```scala
 else params.put("slices", s.toString)
 ```
  as well?

Philippus: I'm not getting a compile error, so I think we're good there.

TactiC (approved):

PR #3171, opened 2024-10-02
Add indices options to update by query

joelvim:
Indices options are available for search API, but not for update API. This PR adds them.


Philippus:
thank you!

PR #3176, opened 2024-10-10
Feat: add support for docvalue_fields and stored_fields in TopHitsAggregation

paulMConstant:
Hello,

It is currently impossible to my understanding to add "docvalue_fields" to a top hits aggregation. Here's a proposition using the storedFields field.

Thanks

### DIFF
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/searches/aggs/TopHitsAggregation.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/searches/aggs/TopHitsAggregation.scala
@@ -16,10 +16,10 @@ case class TopHitsAggregation(name: String,
                               version: Option[Boolean] = None,
                               scripts: Map[String, Script] = Map.empty,
                               storedFields: Seq[String] = Nil,
-                              docValueFields: Seq[String] = Nil,
                               subaggs: Seq[AbstractAggregation] = Nil,
                               metadata: Map[String, AnyRef] = Map.empty,
-                              highlight: Option[Highlight] = None)
+                              highlight: Option[Highlight] = None,
+                              docValueFields: Seq[String] = Nil)
     extends Aggregation {
 
   type T = TopHitsAggregation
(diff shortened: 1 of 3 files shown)


Philippus (changes requested):
Stored fields are not the same thing as doc value fields, see https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-top-hits-aggregation.html. The existing `storedFields` field is not used in the builder, so that could be added. If you want to add `docValueFields` a field would have to be added to the `TopHitsAgrregation` case class and an accompanying `docvalueFields` method similar to what is done in the `InnerHit` case class.

paulMConstant:
Thank you for the quick review. I fixed the `storedFields` addition and added `docValueFields` as done in `InnerHit`.

elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/searches/aggs/TopHitsAggregation.scala
```diff
@@ -16,6 +16,7 @@ case class TopHitsAggregation(name: String,
                               version: Option[Boolean] = None,
                               scripts: Map[String, Script] = Map.empty,
                               storedFields: Seq[String] = Nil,
+                              docValueFields: Seq[String] = Nil,
                               subaggs: Seq[AbstractAggregation] = Nil,
                               metadata: Map[String, AnyRef] = Map.empty,
                               highlight: Option[Highlight] = None)
```
Philippus, line 22: can you move the new field to the end for slightly better backwards compatibility?

paulMConstant: Done

Philippus (approved):

Philippus:
Thank you!

paulMConstant:
Do you think you could make a release soon ?

Philippus:
I'd like to include the update to Elasticsearch 8.15.3, which looks to be imminent.

PR #3209, opened 2024-11-11
You Know, for Search

Philippus:
Applying this PR will add the "serverInfo" GET `/` request, returning version info and the famous tagline. The naming of `serviceInfo`, `MainRequest` and `MainResponse` comes from the Elasticsearch code base, there is no entry in the API documentation about the request.


TactiC (approved):
Nice

PR #3215, opened 2024-11-17
Use http4s.Uri in Http4sClient

lenguyenthanh:
This is to avoid conversion between ElasticNodeEndpoint to http4s.Uri
for every request. Which'll save us some allocation, hence more
performance. This also allows users to pass http4s.Uri directly to
Http4sClient.

### DIFF
--- a/elastic4s-client-http4s/src/main/scala/com/sksamuel/elastic4s/http4s/Http4sClient.scala
+++ b/elastic4s-client-http4s/src/main/scala/com/sksamuel/elastic4s/http4s/Http4sClient.scala
@@ -18,21 +18,6 @@ object Http4sClient {
 
   def usingIO(
     client: http4s.client.Client[IO],
-    endpoint: ElasticNodeEndpoint,
-  )(implicit runtime: IORuntime): Http4sClient[IO] = {
-    val ioRunner = new CallbackRunner[IO] {
-      override def run[A](fa: IO[A], cb: Either[Throwable, A] => Unit): Unit = fa.unsafeRunAsync(cb)
-    }
-
-    Http4sClient(
-      client = client,
-      endpoint = endpoint,
-      runner = ioRunner
-    )
-  }
-
-  def usingIO(
-    client: http4s.client.Client[IO],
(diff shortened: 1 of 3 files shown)


lenguyenthanh:
cc @igor-vovk

elastic4s-client-http4s/src/main/scala/com/sksamuel/elastic4s/http4s/Http4sClient.scala
```diff
@@ -24,18 +24,49 @@ object Http4sClient {
...
     }
 
+    Http4sClient(
+      client = client,
+      endpoint = endpoint,
+      runner = ioRunner
+    )
+  }
+
+  def usingIO(
```
igor-vovk, line 34: Maybe worth totally replace initial `usingIO` method? I don't think someone already using it, it haven't went into the release yet

igor-vovk: Actually I see, it went into `v8.15.4`. Still maybe worth replacing it...

lenguyenthanh: I think this makes sense, did remove the other `usingIO` function.

Philippus:
Let's keep the initial implementation, to be on the safe side. We can deprecate it.

elastic4s-client-http4s/src/main/scala/com/sksamuel/elastic4s/http4s/Http4sClient.scala
```diff
@@ -31,11 +31,43 @@ object Http4sClient {
     )
   }
 
+  @deprecated("Use usingIO with http4s.Uri", "8.15.5")
```
Philippus, line 34:
```suggestion
  @deprecated("Use usingIO with http4s.Uri", "8.16.0")
```

Philippus (approved):

PR #3217, opened 2024-11-19
feat: adding elastic4s-pekko-http-streams module

jmarreroValsea:
Basically copying `elastic4s-http-streams` but with pekko stuff


Philippus:
A bit similar to https://github.com/Philippus/elastic4s/pull/3203, but I'm happy to give you the credit. I'm planning to move some things around before publishing.

PR #3276, opened 2025-01-06
Fix MultiGetResponse error handling

adamseres:
Fixes https://github.com/Philippus/elastic4s/issues/3275.

### DIFF
--- a/elastic4s-handlers/src/main/scala/com/sksamuel/elastic4s/handlers/get/GetHandlers.scala
+++ b/elastic4s-handlers/src/main/scala/com/sksamuel/elastic4s/handlers/get/GetHandlers.scala
@@ -18,13 +18,6 @@ trait GetHandlers {
 
   implicit object MultiGetHandler extends Handler[MultiGetRequest, MultiGetResponse] {
 
-    override def responseHandler: ResponseHandler[MultiGetResponse] = new ResponseHandler[MultiGetResponse] {
-      override def handle(response: HttpResponse): Either[ElasticError, MultiGetResponse] = response.statusCode match {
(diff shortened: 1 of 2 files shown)


elastic4s-handlers/src/main/scala/com/sksamuel/elastic4s/handlers/get/GetHandlers.scala
```diff
@@ -18,13 +18,6 @@ trait GetHandlers {
 
   implicit object MultiGetHandler extends Handler[MultiGetRequest, MultiGetResponse] {
 
-    override def responseHandler: ResponseHandler[MultiGetResponse] = new ResponseHandler[MultiGetResponse] {
-      override def handle(response: HttpResponse): Either[ElasticError, MultiGetResponse] = response.statusCode match {
-        case 404 | 500 => sys.error(response.toString)
-        case _         => Right(ResponseHandler.fromResponse[MultiGetResponse](response))
-      }
-    }
-
```
Philippus, line 27: This will change the semantics for 404 and 500 errors as well. I'd like to be a bit more conservative.
Can you change it to just add an extra case:
```case 401       => Left(ElasticErrorParser.parse(response))``` ?

Philippus (changes requested):

elastic4s-handlers/src/main/scala/com/sksamuel/elastic4s/handlers/get/GetHandlers.scala
Philippus: I'll merge this PR and fix this in another PR.

PR #3308, opened 2025-02-10
Add a handling for the null case when flattening over the map of settings, for example when initial_recovery = null after copying or recovering an index

q-martindurchov:


Philippus:
Is it possible to add a test case?

q-martindurchov:
Unless you are happy with unit tests, which is also an option, just testing the flatten function

q-martindurchov:
Added a test, let me know if there is anything else. I will also implement the CloneAPI hopefully soon. Will update on that when it is ready

Philippus:
Can you format the code? `scalafmtAll` in sbt.

PR #3315, opened 2025-02-17
feat: adds boost support into interval queries

cdmbr:
adds the "[undocumented](https://github.com/elastic/elasticsearch/blob/main/server/src/main/java/org/elasticsearch/index/query/IntervalQueryBuilder.java#L92-L95)" boost into intervals queries


Philippus (approved):

Philippus:
thanks!

PR #3367, opened 2025-03-17
Explicit effect types

igor-vovk:
Superseeds #3119.

Adds explicit effect type into the `ElasticClient` type definition. This simplifies using different effect systems natively, removing double transformation from `F1[_]` into `(Either[Throwable, A]) => Unit` and then into any `F2[_]`. Now http client, `ElasticClient`, and the code using it use the same single `F[_]` type.

Future improvements:
* There is a todo in `docs/effects.md` to sync it with the latest approach, the information there is no longer applicable. Once everyone is okay with those changes, I'm going to update the docs.
* Effect modules now contain only an instance of `Functor`. Maybe they can be merged into the core module, and then cats-effect / zio / etc deps can be added as provided.
* Probably modules like `effect-cats-2`, `effect-zio-1`, `effect-scalaz` (and maybe monix as well) can be removed, I'm not sure anyone uses them these days.

CC @lenguyenthanh

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,14 +190,45 @@ object ArtistIndex extends App {
 ```
 
 
-## Alternative Executors
+## Executors
 
-By default, elastic4s uses scala `Future`s when returning responses, but any effect type can be supported.
+From version 9, elastic4s uses `cats.Functor` typeclass to map effects. This allows to support various effect types for
+all major effect libraries. See below the list of supported libraries and how to work with them.
 
-If you wish to use ZIO, Cats-Effect, Monix or Scalaz, then read this page on [alternative effects](docs/effects.md).
+### Cats Effect IO
 
+1. Add dependency on `cats-effect` library to your project.
+2. Use `import cats.effect.implicits._` to bring implicit instances for `cats.effect.IO`
 
+### Scala Future
 
+To work with `scala.concurrent.Future`, you need to import the following:
+
+```scala
+// 1. bring in the Functor instance for Future:
+import cats.implicits.catsStdInstancesForFuture
+
+// 2. import implicit ExecutionContext
+import scala.concurrent.ExecutionContext.Implicits.global
+```
+
+### ZIO Task
+
+1. Add `elastic4s-effect-zio` dependency to your project.
+2. Use `import com.sksamuel.elastic4s.zio.instances._` to bring implicit instances for `zio.Task`
+
+Alternatively, you can use an official [zio-interop-cats](https://zio.dev/guides/interop/with-cats-effect/) library,
+then `import zio.interop.catz.core._` will bring necessary typeclass implementations.
+
+### Monix Task
+
+1. Add `elastic4s-effect-monix` dependency to your project.
+2. Use `import com.sksamuel.elastic4s.monix.instances._` to bring implicit instances for `monix.eval.Task`
+
+### Scalaz Task
+
+1. Add `elastic4s-effect-scalaz` dependency to your project.
+2. Use `import com.sksamuel.elastic4s.scalaz.instances._` to bring implicit instances for `scalaz.concurrent.Task`
 
 
 ## Index Refreshing
--- a/build.sbt
+++ b/build.sbt
@@ -101,6 +101,9 @@ lazy val scala3Projects: Seq[ProjectReference] = Seq(
   clientsSniffed,
   clientpekko,
   clienthttp4s,
+  zio_1,
+  zio,
+  monix,
   tests,
   testkit,
   circe,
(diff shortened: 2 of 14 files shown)


lenguyenthanh:
thanks @igor-vovk for reviving this, I always keep dreaming about this. I'll look more into it in the next few days.

elastic4s-client-http4s/src/main/scala/com/sksamuel/elastic4s/http4s/Http4sClient.scala
```diff
@@ -1,87 +1,44 @@
...
-    }
-
-    Http4sClient(
-      client = client,
-      endpoint = endpoint,
-      runner = ioRunner
-    )
-  }
-
   def apply[F[_]: Async: Files](
```
lenguyenthanh, line 13: Maybe this is an oversight from last pr, I think we could lower the constraint from `Async` to `Sync` here.

lenguyenthanh (commented):
This is great!

I think we could follow up this with:

- Use cats.Functor instead of our own
- Remove all `effect` module, we could provide some migration path for those who uses these modules.
- Consider removing `ElasticClient`, we could simplify the code base: https://github.com/Philippus/elastic4s/issues/3122#issuecomment-2284660970

I believe with these changes, we'll have a much leaner code base and more perform as well.

elastic4s-client-http4s/src/main/scala/com/sksamuel/elastic4s/http4s/Http4sClient.scala
igor-vovk: Correct, fixed

lenguyenthanh (approved):

igor-vovk:
Regarding the (1) I agree, @lenguyenthanh I don't see what is the need in `effect-*` modules anymore, they are now one-liners and engineers can just implement a Functor for their own Effect system if there is no existing package with cats-integration

igor-vovk:
I see first 2 bullets as a logical work that can be made in this PR as well. @Philippus what do you think if effect modules will go away and will be replaced with `Functor` from `cats`?

Philippus:
> I see first 2 bullets as a logical work that can be made in this PR as well. @Philippus what do you think if `effect-*` modules will go away and will be replaced with `Functor` from `cats`?

I think it would be nice for migration purposes to keep them for at least one release and deprecate those one liners. Is that an option?

igor-vovk:
Hi @Philippus! Sorry for the delay in the answer. I'll then rollback the last commit, and see how to mark them as deprecated. I'll do it in the next few days.

igor-vovk:
Hi @Philippus! In the end I thought the best that can be done is to reuse the `Functor` from `cats` (change introduced in the last commit), and implement existing effect library bridge modules using it.

So:
* `scala.concurrent.Future` implementation, that was implemented alongside `Functor` class, is supported by `cats` out of the box, so I didn't bring it back (it is imported with `import cats.implicits.catsStdInstancesForFuture`, it requires implicit `ExecutionContext` in visibility scope to work). The needed import was added to classes that work with `Future` implementation, for example in tests.
* `cats` modules are still removed, since it works now out-of-the-box for `cats` (there is no dependency on `cats-effect` anymore, since `Functor` is a part of `cats` library, thus they are totally unnecessary now).
* I did bring back `zio`, `zio-1`, `monix`, `scalaz` modules and reimplemented them on top of `cats.Functor`. I didn't add any deprecation remarks, but added in the documentation that anyone who's willing to use `elastic4s` with `zio` can also use https://github.com/zio/interop-cats library instead of `elastic4s-effect-zio` if they would like so (people who had been using libraries that have dependencies to `cats` with `zio` probably have it in the scope already).

Please tell what you think about it.

## Further work:

I think we need to make a table with matrix, showing which http client works best with which IO implementation.

Something like:
* `sttp`: works with `Future`, `cats-effect`, `zio`
* `http4s`: works with `cats-effect`
etc.

I think to open a separate PR to improve the documentation on this.

Philippus:
Awesome work and thank you for your patience.

Philippus:
I've released version [9.0.0-RC1](https://github.com/Philippus/elastic4s/releases/tag/v9.0.0-RC1). Please take it for a spin!

lenguyenthanh:
This is a tremendous improvement: https://github.com/lichess-org/lila-search/pull/485

thanks @igor-vovk and @Philippus 🙏

igor-vovk:
I've made a proof that it will work for us in ZD where we use this library extensively. We use OpenSearch's http client which doesn't have effects support out of the box (as opposed to http4s and sttp clients), but still this change simplifies the integration with it.

lenguyenthanh:
I also want to confirm that, this been working well in lichess prod for awhile.

PR #3385, opened 2025-04-04
fix: boosting query parameters

mcenkar:
- Removed not used `queryName` field,
- Removed incorrect `boost` field,
- Made required `negativeBoost` field.

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-boosting-query.html#boosting-top-level-params

Previously it was possible to create boosting query with incorrect body, for example with field `boost` that doesn't exist, or without required `negativeBoost`.


mcenkar:
Hi @Philippus, one test is failing (ScrollPublisherVerificationTest) but it doesn't seem related to my PR, and it was passing before my merge-update today

Philippus (changes requested):
Can you rename the commit and the PR? It's not really a "refactor", it's more a "fix".

mcenkar:
Sure, done, there are some failing tests but not from me

Philippus:
I'll add a PR on top of this PR to facilitate with migration.

PR #3449, opened 2025-05-15
Build and release with java 17

Philippus:


TactiC (approved):

Philippus:
Yes, a lot of Scala libraries are moving away from java 8, and skipping java 11. I'd like to do the same.

PR #3498, opened 2025-06-13
feat: supports knn as query

cdmbr:
supports knn as [query](https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-knn-query) too


Philippus (approved):
LGTM

PR #3538, opened 2025-07-10
feat: added matched queries to inner hit

jfhetzer:
Added `matchedQueries` to the `InnerHit`. Matched queries cannot just exist in the top level `SearchHit` but also in [Inner Hits](https://www.elastic.co/docs/reference/elasticsearch/rest-apis/retrieve-inner-hits) when having a [Named Query](https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-bool-query#named-queries) in a [nested structure](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/nested). Until now it was not possible to retrieve the matched queries from an inner hit.


Philippus:
Would it be possible for you to add a test case for this situation?

jfhetzer:
Hey @Philippus any updates on this one? Is something blocking it from getting merged?

PR #3546, opened 2025-07-17
Elastic index lifecycle management politics support

UralRoman:
Implemented DSL for Index Lifecycle management politcs

### DIFF
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/ElasticDsl.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/ElasticDsl.scala
@@ -17,7 +17,7 @@ import com.sksamuel.elastic4s.handlers.index.{
   IndexTemplateHandlers,
   RolloverHandlers
 }
-import com.sksamuel.elastic4s.handlers.indexlifecyclemanagement.IndexLifecycleManagementHandlers
+import com.sksamuel.elastic4s.handlers.indexlifecyclemanagement.IndexLifecycleManagmentHandlers
 import com.sksamuel.elastic4s.handlers.locks.LocksHandlers
 import com.sksamuel.elastic4s.handlers.main.MainHandlers
 import com.sksamuel.elastic4s.handlers.nodes.NodesHandlers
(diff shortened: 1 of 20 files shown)


apavlychev:
@Philippus Are there any problems to merge this PR ?

elastic4s-core/src/main/scala/com/sksamuel/elastic4s/ElasticDsl.scala
```diff
@@ -17,7 +17,7 @@ import com.sksamuel.elastic4s.handlers.index.{
   IndexTemplateHandlers,
   RolloverHandlers
 }
-import com.sksamuel.elastic4s.handlers.indexlifecyclemanagement.IndexLifecycleManagementHandlers
+import com.sksamuel.elastic4s.handlers.indexlifecyclemanagement.IndexLifecycleManagmentHandlers
```
Philippus: `IndexLifecycleManagementHandlers` is the correct spelling.

elastic4s-domain/src/main/scala/com/sksamuel/elastic4s/requests/indexlifecyclemanagement/policy/IndexLifecyclePolicy.scala
```diff
@@ -0,0 +1,49 @@
...
+
+import com.fasterxml.jackson.databind.JsonNode
+import com.sksamuel.elastic4s.JacksonSupport
+import com.sksamuel.elastic4s.json.{JsonValue, StringValue}
+
+import scala.collection.JavaConverters._
+
+case class IndexLifecyclePolicy(
+    name: String,
+    phases: List[IndexLifecyclePolicyPhase],
```
Philippus, line 11: I think maybe a `Map` is more appropriate here? Then you don't need to override the equals method.

Philippus (changes requested):
Could you rebase this on the current default branch?

elastic4s-domain/src/main/scala/com/sksamuel/elastic4s/requests/indexlifecyclemanagement/policy/IndexLifecyclePolicy.scala
UralRoman: The idea of this solution is that the user can use only the phases that exist in the API. The user should use the phase enumeration rather than create the phases themselves. Therefore, Map will not help here.

Philippus: ok, then how about making it a `Set` instead?

PR #3620, opened 2025-09-09
Fix double encoding path for http4s client

skuayer:
When using http4s client I found that id is stored incorrectly in ElasticSearch in case then it contains special characters. The trouble was in final resulting  URL.  

`ElasticUrlEncoder` in `IndexHandler` encoded path and convertor `org.http4s.Segment.apply` in `RequestResponseConverters. elasticRequestToHttp4sRequest` also encoded path before sending request.


Philippus (approved):

Philippus:
@igor-vovk Do you have some time to take a look? I think you use http4s.

igor-vovk:
Unfortunately I can't test it since I'm not using it anymore in my own projects, but the PR is LGTM 👍

PR #3649, opened 2025-10-03
Add support for preserve_original parameter of the edge n-gram token filter

dragisak:
See https://www.elastic.co/docs/reference/text-analysis/analysis-edgengram-tokenfilter#analysis-edgengram-tokenfilter-configure-parms


Philippus:
Could you add a test case to the `EdgeNGramTokenFilterTest` ?

dragisak:
I intentionally didn't make the change in the deprecated code. So added checks `TokenFilterTest`

Philippus:
thanks!

PR #3650, opened 2025-10-03
Fix inconsistent parameter naming. Snake case -> camel case

dragisak:


elastic4s-domain/src/main/scala/com/sksamuel/elastic4s/analysis/TokenFilter.scala
```diff
@@ -96,7 +96,7 @@ case class WordDelimiterGraphTokenFilter(
   def protectedWords(words: String): WordDelimiterGraphTokenFilter       = copy(protectedWords = words.some)
   def protectedWordsPath(path: String): WordDelimiterGraphTokenFilter    = copy(protectedWordsPath = path.some)
   def splitOnNumerics(bool: Boolean): WordDelimiterGraphTokenFilter      = copy(splitOnNumerics = bool.some)
-  def stemEnglishPossesive(bool: Boolean): WordDelimiterGraphTokenFilter = copy(stem_english_possessive = bool.some)
+  def stemEnglishPossesive(bool: Boolean): WordDelimiterGraphTokenFilter = copy(stemEnglishPossessive = bool.some)
```
Philippus: It would be neat to deprecate this method and fix the typo in a new `stemEnglishPossessive` method.

Philippus (changes requested):
Good catch! We would need a deprecation path though, for folks using the complete constructor. Something like I did here: https://github.com/Philippus/elastic4s/pull/3653.

dragisak:
Hm. now apply becomes ambiguous. The parameter list has the same number of parameters of the same type.

Philippus (approved):

Philippus:
> Hm. now apply becomes ambiguous. The parameter list has the same number of parameters of the same type.

Ah yes, that is true.

PR #3714, opened 2025-11-06
Add bool filter constructor to QueryApi

ChrisMcD1:
Hello,

First PR here, so sorry if I've broken any cultural norms.

I was just reading the elasticsearch docs, and saw that `filter` was an option I was unfamiliar with. When I went to try to use it in the QueryApi, I couldn't find it. I then thought maybe it nedeed to be added to BoolQuery itself, but evidently it just was not present in the helpers. Given that it could be a drop-in replacement for `must` in many circumstances, I think it is worthy of a helper as well.


Philippus:
Thank you!

PR #3764, opened 2025-12-23
Fix scheme detection in http4s client

skuayer:
For http4s client scheme was build from prefix instead of protocol information.


Philippus (approved):
Nice catch!

PR #3800, opened 2026-02-04
Add CommonRequestOptions to constructor of ScrollPublisher and SearchIterator

ChrisMcD1:
It seems that currently there is no way to make the ScrollPublisher include authentication in its request to elasticsearch, since it calls the `execute` method internally, but has no way to pass in the `CommonRequestOptions` that would contain that authentication.

Given that the `ScrollPublisher` has a singular `SearchRequest` in its construction, I think it makes sense to include the options for that request alongside it.

I noticed the same thing in `SearchIterator`, and added it to

I also searched around the code base, and it seems that the following other classes have calls to `ElasticClient.execute`, and no way to pass options to them. However, they do not have their request object taken in at construction, and I am not currently using any of them, so I wasn't sure what the best way to pass in the options to them would be.
* `BulkIndexingSubscriber`
* `BatchElasticSink`
* `ElasticSource`

---

If anyone has a workaround to get authentication working with these at all in the interim, I would love to hear it! Currently planning to vendor this dependency with these patches.

### DIFF
--- a/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/searches/SearchIterator.scala
+++ b/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/requests/searches/SearchIterator.scala
@@ -1,6 +1,6 @@
 package com.sksamuel.elastic4s.requests.searches
 
-import com.sksamuel.elastic4s.{ElasticClient, HitReader, RequestFailure, RequestSuccess}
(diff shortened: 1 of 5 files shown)


ChrisMcD1:
Also, this will need to be integrated with https://github.com/Philippus/elastic4s/pull/3724, depending on who merges first

Philippus (approved):

Philippus:
I'll create a PR for the other three cases.

PR #3858, opened 2026-03-17
Support parent task id in tasks API

joelvim:


Philippus:
I see you targeted the 8.x branch, was that intentional?

joelvim:
Yes, sorry, I should have put a comment. We currently use the 8.x version of elasticsearch, so we need it for at least this version, but maybe it is better to target the 9.x and to backport it to the 8.x?

Philippus:
No problem, it's fine like this. If you could make the same change targeting 9.x that would be great!
And, since you need it, I'll cut a release this week.

Philippus (approved):

joelvim:
Thank you, I did another PR against series/9.x to port the same change. I let you deal with the rest of the process, thanks a lot for your work!

Philippus:
https://github.com/Philippus/elastic4s/releases/tag/v8.19.1 should now, or soon, be available.

PR #3887, opened 2026-03-27
feat: add shard preference support for update by query async requests

flavienbert:
Add support for the `preference=_shards:0,1,2,3` parameter in async update-by-query operations through a new ShardsPreferenceRequest case class.

Changes:
- Add ShardsPreferenceRequest case class to model shard preferences. The goal of this model is to have a nonEmptyList
- Add `shards: Option[ShardsPreferenceRequest]` field to UpdateByQueryAsyncRequest
- Add fluent API methods: shards(Int) and shards(Set[Int])
- Propagate shard preference parameter to HTTP request in UpdateByQueryHandler
- Add comprehensive unit tests validating preference parameter building


flavienbert:
wrong target

PR #3889, opened 2026-03-27
[series/9.x] add shard preference support for update by query async requests

flavienbert:
Add support for the `preference=_shards:0,1,2,3` parameter in async update-by-query operations through a new ShardsPreferenceRequest case class.

Changes:

- Add ShardsPreferenceRequest case class to model shard preferences. The goal of this model is to have a nonEmptyList
- Add `shards: Option[ShardsPreferenceRequest]` field to UpdateByQueryAsyncRequest
- Add fluent API methods: shards(Int) and shards(Set[Int])
- Propagate shard preference parameter to HTTP request in UpdateByQueryHandler
- Add comprehensive unit tests validating preference parameter building

### DIFF
--- a/elastic4s-core/src/test/scala/com/sksamuel/elastic4s/requests/update/UpdateByQueryHandlerTest.scala
+++ b/elastic4s-core/src/test/scala/com/sksamuel/elastic4s/requests/update/UpdateByQueryHandlerTest.scala
@@ -1,6 +1,7 @@
 package com.sksamuel.elastic4s.requests.update
 
 import com.sksamuel.elastic4s.handlers.update.UpdateHandlers
+import com.sksamuel.elastic4s.requests.common.Preference.Shards
 import com.sksamuel.elastic4s.requests.searches.queries.matches.MatchAllQuery
 import org.scalatest.flatspec.AnyFlatSpec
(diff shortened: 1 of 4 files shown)


Philippus:
As far as I can tell, shard preference isn't the only preference option available for that call. I'd prefer to model it the same as we do with GetRequest, as in:
```
  def preference(pref: Preference): GetRequest = preference(pref.value)
  def preference(pref: String): GetRequest = copy(preference = pref.some)
```

flavienbert:
The modification is done. 
I initially overlooked the existing model, but I’m not a big fan of it:
`case class Shards(ids: List[String]) extends Preference("_shards:" + ids.mkString(","))`.

It doesn’t enforce that at least one shard is provided or same shards are provided, and shards are defined as String instead of Int.
My version `case class ShardsPreferenceRequest(shard: Int, optionalShards: Option[Set[Int]] = None)` feels safer and more type-safe.

Philippus (approved):

Philippus (approved):

Philippus:
I agree improvements could be made to the preferences, it's also possible to combine shards with other preferences (https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-search#operation-search-preference).

PR #3918, opened 2026-04-14
[series/9.x] add response field to GetTaskResponse

flavienbert:
Introduced a new optional `response` field to the GetTaskResponse case class to support additional task response details.


Philippus:
Would it be possible to add one or more tests?

Philippus:
@flavienbert are you planning to add some tests?

